PR #13, opened 2021-08-25
solution

SashaRudenko:
[DEMO LINK](https://SashaRudenko.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import PropsTypes from 'prop-types'
+import './Person.scss';
+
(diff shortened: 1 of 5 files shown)


src/components/Person.jsx
```diff
@@ -0,0 +1,45 @@
...
+      <p className="Person__partner">
+        My husband&apos;s name is {partnerName}
+      </p>
+    )}
+    {isMarried 
+      && sex === 'm' 
+      && (
+        <p className="Person__partner">
+          My wife&apos;s name is {partnerName}
+        </p>
```
nastyyvolk, line 28: There is a way to handle this using one conditional structure

nastyyvolk (approved):
Nice!

PR #20, opened 2021-08-29
Person component created

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/react_article/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.scss';
 
+import { Person } from './components/Person';
+
 const misha = {
   name: 'Misha',
   age: 37,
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -25,22 +27,27 @@ const alex = {
...
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person
+      name={misha.name}
+      age={misha.age}
+      sex={misha.sex}
+      isMarried={misha.isMarried}
+      partnerName={misha.partnerName}
+    />
```
DmytroMykoliv, line 36: You can use destructuring here and in the following:
```suggestion
    <Person {...misha} />
```

DmytroMykoliv (approved):
Greate work )

src/App.js
varseniuk: Dmytro is right. Let's use destructuring

src/components/Person.js
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <section className="Person">
+
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      <p className="Person__age">
+        {age && `I am ${age}`}
+      </p>
```
varseniuk, line 25:
```suggestion
      {age && (
        <p className="Person__age">
          I am {age}
        </p>
      )}
```

Noe you will render an empty `p` if there is no `age`. Let's not render it at all

varseniuk (changes requested):
All is good but let's make small fixes

src/components/Person.js
ihor-jpeg: thank you for noticing this mistake. i have fixed it already

src/App.js
ihor-jpeg: great idea! thank you

src/components/Person.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <section className="Person">
+
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age
+        && (
```
varseniuk, line 26:
```suggestion
      {age && (
```
it will be better formatting

varseniuk (approved):
🔥

varseniuk (changes requested):
🔥

varseniuk (approved):

PR #52, opened 2021-09-01
Solution Person

AndriyNikorych:
[DEMO LINK](https://AndriyNikorych.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+}) => {
+  const sexOfPartner = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      <p className="Person__age">{age ? `I am ${age}` : ''}</p>
+      <p className="Person__partner">
+        {isMarried
+          ? `My ${sexOfPartner}\`s name is ${partnerName}`
+          : 'I am not married'
+        }
+      </p>
+    </section>
+  );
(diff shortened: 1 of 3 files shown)


src/components/Person.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
```
Clonvhs, line 9: you can also add you func params in to  the consts like  let {a,b,c} = props

Clonvhs (approved):
Great one

src/components/Person.jsx
```diff
@@ -0,0 +1,40 @@
...
+}) => {
+  const sexOfPartner = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      <p className="Person__age">{age ? `I am ${age}` : ''}</p>
```
varseniuk, line 18:
```suggestion
      {age && (
        <p className="Person__age">
          I am {age}
        </p>
      )}
```
Now you are rendering an empty `p` if there's no `age` value. Let's not render it at all in this case

varseniuk (changes requested):
All is good, but let's fix the conditional rendering

src/components/Person.jsx
```diff
@@ -0,0 +1,45 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+        <p className="Person__age">
+          {`I am `}
+          {age}
```
varseniuk, line 21:
```suggestion
          I am {age}
```

varseniuk (approved):

PR #72, opened 2021-10-07
solution

s-person27:
[DEMO LINK](https://s-person27.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -13,13 +13,19 @@ export const Person = ({
     <h2 className="Person__name">
       {`My name is ${name}`}
     </h2>
-    {age !== 0 ? (
+    {/* {age !== 0 ? (
       <p className="Person__age">
         {`I am ${age}`}
       </p>
     ) : (
       null
+    )} */}
+    {age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
     )}
+
     {isMarried ? (
       <p className="Person__partner">
         {'My '}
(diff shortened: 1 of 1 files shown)


src/components/Person.jsx
```diff
@@ -0,0 +1,52 @@
...
+    <h2 className="Person__name">
+      {`My name is ${name}`}
+    </h2>
+    {age !== 0 ? (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+    ) : (
+      null
+    )}
```
aliceblunt8, line 22: use logical operator (&&)

s-person27: I see '0' in the browser instead of just an empty line when I use && and a default value of age is 0. So I change a default value of age to null and add It to propTypes. I'm not sure that I'm doing It correctly

src/components/Person.jsx
```diff
@@ -0,0 +1,58 @@
...
+        {`I am ${age}`}
+      </p>
+    ) : (
+      null
+    )} */}
+    {age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+    )}
```
Anna-Tkachenko, line 27: 🔥
Delete redundant comments)

src/components/Person.jsx
```diff
@@ -0,0 +1,58 @@
...
+          ? 'wife\'s'
+          : 'husband\'s'
+        }
+        {` name is ${partnerName}`}
+      </p>
+    ) : (
+      'I am not married'
+    )
+
+      }
```
Anna-Tkachenko, line 42: It's better to create separate variable before return statement where you can store the result of inner conditional statement (checking sex === 'm')
```suggestion
    {isMarried ? (
      <p className="Person__partner">
        {`My ${partner}'s name is ${partnerName}`}
      </p>
    ) : (
      'I am not married'
    )

      }
```

PR #92, opened 2021-10-08
Develop

Saharij:
[DEMO](https://Saharij.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import './Person.scss';
+import PropTypes from 'prop-types';
+
+export const Person = ({ name, age, sex, isMarried, partnerName }) => (
+  <section className="Person">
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -0,0 +1,36 @@
...
+    {age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+    )}
+    <p className="Person__partner">
+      {isMarried
+        ? `My ${sex === 'm' ? 'wife' : 'husband'}'s name is ${partnerName}`
+        : 'I am not married'
+      }
```
Anna-Tkachenko, line 19: It's better to create separate variable before return statement where you can store the result of inner checking (sex === 'm')
```suggestion
      {isMarried
        ? `My ${partner}'s name is ${partnerName}`
        : 'I am not married'
      }
```

PR #100, opened 2021-10-10
react_person PR

Rocketjet:
[LINK](https://Rocketjet.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import './Person.scss';
+
+const Person = ({
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+}) => (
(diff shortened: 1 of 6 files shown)


src/components/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+    {age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {isMarried
+        ? `My ${sex === 'm' ? "wife's" : "husband's"} name is ${partnerName}`
+        : `I am not married`}
```
Anna-Tkachenko, line 26: It's better to create separate variable before return statement where you can store the result of inner conditional statement (checking sex === 'm')
```suggestion
      {isMarried
        ? `My ${partner}'s name is ${partnerName}`
        : `I am not married`}
```

PR #105, opened 2021-10-11
solution

caelum29:
[DEMO LINK](https://caelum29.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -1,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Person = ({person}) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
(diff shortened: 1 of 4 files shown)


src/components/Person.jsx
```diff
@@ -0,0 +1,42 @@
...
+    isMarried,
+    partnerName,
+  } = person;
+  const partner = sex === 'm'
+    ? `My wife's name is ${partnerName}`
+    : `My husband's name is' ${partnerName}`
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      <p className="Person__age">{age && `I am ${age}`}</p>
```
Anna-Tkachenko, line 18: This paragraph shouldn't render if age is null

Anna-Tkachenko (approved):

PR #153, opened 2021-12-03
completed react person

i-stanko:


lanebx:
good job)
but i think i need to change the logic for "age"

Here:
Снимок экрана 2021-12-03 в 13 44 56

lanebx (approved):

i-stanko:
> good job) but i think i need to change the logic for "age"
> 
> Here: Снимок экрана 2021-12-03 в 13 44 56

it was intended that it is indecent to ask girls about their age

src/App.js
```diff
@@ -25,22 +26,9 @@ const alex = {
...
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person {...misha} />
+    <Person {...olya} />
+    <Person {...alex} />
```
aionova, line 31: you could do it inside of map

aionova (approved):

PR #171, opened 2021-12-05
solution

Nadiyahr:


src/components/Person.js
```diff
@@ -0,0 +1,43 @@
...
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {' '}
+      {name}
+    </h2>
+    <p className="Person__age">
+      {age && `I am ${age}`}
+    </p>
+    {isMarried ? (
```
WorkAccCom, line 15:
```suggestion
    {isMarried 
    ? (
```

WorkAccCom (approved):
Well done!

src/components/Person.js
```diff
@@ -0,0 +1,45 @@
...
+  </section>
+);
+
+Person.defaultProps = {
+  age: null,
+  sex: '',
+  isMarried: false,
+};
+
+Person.propTypes = {
```
AlinaMatuschak, line 40: Should add prop type for partnerName

src/components/Person.js
```diff
@@ -0,0 +1,45 @@
...
+          I am single
+        </p>
+      )}
+  </section>
+);
+
+Person.defaultProps = {
+  age: null,
+  sex: '',
+  isMarried: false,
```
AlinaMatuschak, line 37: What if we get isMarried true but don't get sex at all?

src/components/Person.js
```diff
@@ -0,0 +1,45 @@
...
+    <p className="Person__age">
+      {age && `I am ${age}`}
+    </p>
+    {isMarried
+      ? (
+        <p className="Person__partner">
+          {
+              sex === 'm'
+                ? 'My wife'
+                : 'My husband'
```
AlinaMatuschak, line 21: Can we create a variable for that?

AlinaMatuschak (changes requested):

src/components/Person.js
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {' '}
+        {name}
+      </h2>
+      <p className="Person__age">
+        {age && `I am ${age}`}
+      </p>
```
AlinaMatuschak, line 17: if person don't have age, don't render Person__age at all

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #181, opened 2021-12-06
Solution

Vitalii-Lytvyn01:
[DEMO LINK](https://Vitalii-Lytvyn01.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import './Person.scss';
+
+export function Person({ name, sex, age, isMarried, partnerName }) {
+  const partner = sex === 'm' ? `wife's` : `husband's`;
+
+  return (
+    <>
+      <h2 className="Person__name">
+        {'My name is '}
+        {name}
+      </h2>
+      <p className="Person__age">
+        {'I am '}
+        {age}
+      </p>
+      <p className="Person__partner">
+        {isMarried
+          ? `My ${partner} name is ${partnerName}`
+          : 'I am not married'}
+      </p>
+    </>
+  );
+}
+
+Person.defaultProps = {
+  name: 'User',
+  age: 999,
+  sex: 'Undefined',
+  isMarried: false,
+};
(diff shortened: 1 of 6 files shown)


src/components/Person.jsx
```diff
@@ -0,0 +1,31 @@
...
+          ? `My ${partner} name is ${partnerName}`
+          : 'I am not married'}
+      </p>
+    </>
+  );
+}
+
+Person.defaultProps = {
+  name: 'User',
+  age: 999,
```
DudarDaryna, line 28: if there isn't information about age, u should just skip it)

DudarDaryna (approved):
GJ

src/components/Person.jsx
```diff
@@ -0,0 +1,31 @@
...
+      <p className="Person__partner">
+        {isMarried
+          ? `My ${partner} name is ${partnerName}`
+          : 'I am not married'}
+      </p>
+    </>
+  );
+}
+
+Person.defaultProps = {
```
AlinaMatuschak, line 26: use PropTypes with defaultProps

AlinaMatuschak (changes requested):
fix @DudarDaryna comment

src/components/Person.jsx
Vitalii-Lytvyn01: As i understand react needs either isRequired for a property or a default setting .

AlinaMatuschak: need add expected types for props (all props). Some props can be required, but if not, they must have a default value

src/components/Person.jsx
```diff
@@ -0,0 +1,41 @@
...
+  age: PropTypes.number,
+  isMarried: PropTypes.bool,
+  partnerName: PropTypes.string,
+};
+
+Person.defaultProps = {
+  name: 'User',
+  sex: 'Undefined',
+  isMarried: false,
+  age: 0,
```
AlinaMatuschak, line 39: better don't render age if we don't have this value

AlinaMatuschak (approved):

PR #246, opened 2022-02-02
Solution

marinagera:
https://marinagera.github.io/react_person/


src/App.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import './App.scss';
+import Person from './components/Person';
```
vkryvytskyy, line 2: Этого компонента нет в pull request, добавь его

vkryvytskyy (changes requested):

src/components/Person.jsx
```diff
@@ -0,0 +1,32 @@
...
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {' '}
+        {name}
+      </h2>
+      <p className="Person__age">
+        {age && (`I am ${age}`)}
+      </p>
```
aliceblunt8, line 15: better to use conditional rendering for the whole p element with text

src/components/Person.jsx
```diff
@@ -0,0 +1,32 @@
...
+      <p className="Person__partner">
+        {!isMarried ? 'I am not married'
+          : `My
+        ${isMarried && ((sex === 'm' && 'wife')
```
aliceblunt8, line 19: why do you need to check isMarried here too? you've already checked it in ternary

aliceblunt8 (changes requested):

danylo-onishchuk (approved):

PR #253, opened 2022-02-10
Develop

Oleksii-Pinchuk:
[DEMO LINK](https://Oleksii-Pinchuk.github.io/react_person/)

### DIFF
--- a/src/Components/Person/Person.jsx
+++ b/src/Components/Person/Person.jsx
@@ -10,12 +10,9 @@ export const Person = ({
   partnerName,
 }) => {
   const spouse = (sex === 'm' ? 'wife' : 'husband');
-  const personPartner = (isMarried
-    ? (
-      <p className="Person__partner">
-        {`My ${spouse}'s name is ${partnerName}`}
-      </p>
-    ) : (<p className="Person__partner">I am not married</p>));
+  const personPartner = isMarried
+    ? `My ${spouse}'s name is ${partnerName}`
+    : 'I am not married';
 
   return (
     <section className="Person">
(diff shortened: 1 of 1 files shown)


GitRoko (approved):
Все ок!

src/Components/Person/Person.jsx
```diff
@@ -0,0 +1,32 @@
...
+    )}
+
+    {isMarried ? (
+      <p className="Person__partner">
+        My
+        {` ${sex === 'm' ? 'wife' : 'husband'}'s `}
+        name is
+        {` ${partnerName}`}
+      </p>
+    ) : (
```
Anna-Tkachenko, line 28: It's hard to understand this code with many conditional statements
Let's create variable for storing the result of this check and then use it

Anna-Tkachenko (changes requested):

src/Components/Person/Person.jsx
```diff
@@ -0,0 +1,33 @@
...
+  isMarried,
+  partnerName,
+}) => {
+  const spouse = (sex === 'm' ? 'wife' : 'husband');
+  const personPartner = (isMarried
+    ? (
+      <p className="Person__partner">
+        {`My ${spouse}'s name is ${partnerName}`}
+      </p>
+    ) : (<p className="Person__partner">I am not married</p>));
```
Anna-Tkachenko, line 18: Don't store all tag in variable, only string value that should be in this tag

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job 👍

PR #256, opened 2022-02-11
sol

Kengat:
[DEMO LINK](https://Kengat.github.io/react_person/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # React person
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_person/)
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://Kengat.github.io/react_person/)
 - Follow the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline)
 
 ## Task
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import './Person.scss';
+
+const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = { ...person };
```
polosanya, line 6: Деструктуризацию лучше делать сразу когда принимаешь параметры

polosanya (approved):
Good job!

src/components/Person/Person.jsx
Anna-Tkachenko: @polosanya this way is better)

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,25 @@
...
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age ? <p className="Person__age">{`I am ${age}`}</p> : null}
+      {!isMarried
+        ? <p className="Person__partner">I am not married</p>
+        : (
+          <p className="Person__partner">
+            {`My ${partnerSex}'s name is ${partnerName}`}
+          </p>
+        )
```
Anna-Tkachenko, line 19: Don't repeat yourself: <p className="Person__partner"> repeated in your code
It's better to create one more variable to store string, that depends on isMarried property and use it in this paragraph

Anna-Tkachenko (approved):

PR #276, opened 2022-02-11
person v1

Oleksii-Fedorenko:
[DEMO LINK](https://Oleksii-Fedorenko.github.io/react_person/)

### DIFF
--- a/src/Person/Person.jsx
+++ b/src/Person/Person.jsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-expressions */
 import React from 'react';
 
 import './Person.scss';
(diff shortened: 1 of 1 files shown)


RomaBoreiko (approved):
Looks cleaver!)

src/Person/Person.jsx
```diff
@@ -0,0 +1,47 @@
...
+  isMarried,
+  partnerName,
+}) {
+  let spouse;
+
+  if (sex === 'm') {
+    spouse = 'wife';
+  } else if (sex === 'f') {
+    spouse = 'husband';
+  }
```
Anna-Tkachenko, line 18: It's better to use ternary operator if you should define variable value depending on some check

Anna-Tkachenko (approved):
Good job 👍

PR #379, opened 2022-05-25
solutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_person/)


src/components/Person.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import './Person.scss';
+
```
ZUBOV-ILLIA, line 3: need to import also prop-types

ZUBOV-ILLIA (changes requested):
Good job!
Just add prop-types more

src/components/Person.jsx
```diff
@@ -0,0 +1,35 @@
...
+function marriage(props) {
+  if ((props.props.sex === 'm') && (props.props.isMarried)) {
+    return (`My wife's name is ${props.props.partnerName}`);
+  }
+
+  if ((props.props.sex === 'f') && (props.props.isMarried)) {
+    return (`My husband's name is ${props.props.partnerName}`);
+  }
+
+  return ('I am not married');
```
Anna-Tkachenko, line 13: Don't repeat yourself, you change only one word in two if conditions
```suggestion
  const partnerSex = sex === 'm' ? 'wife' : 'husband';
  
  if (isMarried) {
    return (`My ${partnerSex}'s name is ${partnerName}`);
  }
  
  return ('I am not married');
```

Anna-Tkachenko: Also use destructuring for getting main variables
And check the name of the function, it should be understandable, for example: getMarriageStatus

src/components/Person.jsx
```diff
@@ -0,0 +1,35 @@
...
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {` ${props.props.name}`}
+      </h2>
+      <p className="Person__age">
+        {
+          props.props.age ? (`I am  ${props.props.age}`) : ''
+        }
+      </p>
```
Anna-Tkachenko, line 27: You shouldn't render this tag if age is null
{age && (
	<p> ... </p>
)}

src/components/Person.jsx
```diff
@@ -0,0 +1,35 @@
...
+      <p className="Person__parthner">
+        {marriage(props)}
```
Anna-Tkachenko, line 29: Will be better to create separate variable before return statement, save here the result of this function and use this variable in jsx

src/components/Person.jsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  if ((props.props.sex === 'f') && (props.props.isMarried)) {
+    return (`My husband's name is ${props.props.partnerName}`);
+  }
+
+  return ('I am not married');
+}
+
+function Person(props) {
```
Anna-Tkachenko, line 16: Use desctructuring for getting main values from props object

src/components/Person.jsx
Anna-Tkachenko: agree, add prop types

Anna-Tkachenko (changes requested):

src/components/Person.jsx
```diff
@@ -0,0 +1,57 @@
...
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {` ${name}`}
+      </h2>
+      <p className="Person__age">
+        {
+          age ? (`I am  ${age}`) : ''
+        }
+      </p>
```
Anna-Tkachenko, line 32: Don't render this tag if age is null
{age && (
  <p> ... </p>
)}

Anna-Tkachenko (approved):

PR #381, opened 2022-05-25
First try

RomekIvantsiv:
[DEMO LINK](https://RomekIvantsiv.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -6,7 +6,7 @@ export const Person = ({ person }) => {
     name, age, sex, isMarried, partnerName,
   } = person;
 
-  const personPartner = sex === 'f' ? 'wife' : 'husband';
+  const personPartner = sex === 'm' ? 'wife' : 'husband';
 
   return (
 


ZUBOV-ILLIA (changes requested):
Good Job!!!
Ned add just one small prop-types )

src/components/Person.jsx
```diff
@@ -0,0 +1,70 @@
...
+  isMarried: false,
+};
+
+export const Person = ({ person }) => {
+  const {
+    name, age, sex, isMarried, partnerName,
+  } = person;
+
+  return (
+    <>
```
Denchepornyuk, line 34: Do you need `React.Fragment` here?

src/components/Person.jsx
```diff
@@ -0,0 +1,70 @@
...
+        </p>
+
+        { isMarried
+          ? (
+            <p className="Person__partner">
+              { sex === 'm'
+                ? `My wife's name is ${partnerName}`
+                : `My husband's name is ${partnerName}`}
+            </p>
+          )
```
Denchepornyuk, line 51: don't use a ternary operator inside another ternary operator

Denchepornyuk (changes requested):

src/components/Person.jsx
```diff
@@ -0,0 +1,71 @@
...
+  partnerName: 'Maksym',
+};
+
+// eslint-disable-next-line
+export const alex = {
+  name: 'Alex',
+  age: 25,
+  sex: 'm',
+  isMarried: false,
+};
```
Anna-Tkachenko, line 26: You shouldn't add this objects in Person.jsx file, this data should be passed in props for the Person component

src/components/Person.jsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      <p className="Person__age">
+        { age
+          ? `I am ${age}`
+          : null}
+      </p>
```
Anna-Tkachenko, line 48: Don't render this paragraph if there is no age
```suggestion

        { age
          ? <p> `I am ${age}` </p>
          : null}
```

src/components/Person.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+      <p className="Person__partner">
+        {isMarried
+          ? (`My ${partner}'s name  is ${partnerName}`) : ('I am not married') }
+      </p>
```
Anna-Tkachenko, line 53: Will be better to create separate variable before the return statement with final married status message and use this variable here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (changes requested):
You didn't add any commits after my comments

Anna-Tkachenko (approved):

PR #389, opened 2022-05-26
Solved 'React person' task

dellachaise:
[DEMO LINK](https://dellachaise.github.io/react_person/)


Krooss (approved):

lytvyshko (approved):

src/components/Person.jsx
```diff
@@ -0,0 +1,39 @@
...
+    partnerName,
+  } = person;
+
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      <p className="Person__age">{age && `I am ${age}`}</p>
+      {isMarried ? <p className="Person__partner">{`My ${partner}'s name is ${partnerName}`}</p> : 'I am not married'}
```
Anna-Tkachenko, line 20: You should render some string inside p tag
Now the last part of the ternary operator is rendered without any tag
Also will be better to compute married status string before return statement in your component
Such way can help to increase understanding your code and make it more clear
```suggestion
      <p ...>
        {isMarried
          ? {marriedStatus}
          : 'I am not married'
        }
      </p>
```

src/components/Person.jsx
```diff
@@ -0,0 +1,39 @@
...
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  person: {
```
Anna-Tkachenko, line 26: You should add default props for all non required fields in person object
Or make other fields required except the name

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #390, opened 2022-05-26
peopel`s info in cards

AndriiRut:
[DEMO LINK](https://AndriiRut.github.io/react_person/)


src/componenta/personal_info/index.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import './person.scss';
```
kitaminary, line 2: use index.jsx just for exports

kitaminary (changes requested):
you demo-link doesn't work

Anna-Tkachenko (changes requested):
You didn't add PersonalInfo component files

src/componenta/personalInfo/personalInfo.jsx
```diff
@@ -0,0 +1,32 @@
...
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
+        ) : null}
+      <p className="Person__partner">
+        {MarriedOrNot(
+          sex,
+          isMarried,
+          partnerName,
+        )}
```
Anna-Tkachenko, line 28: Will be better to create separate variable and store here the result of this function
And use this variable in jsx

Anna-Tkachenko (approved):

PR #392, opened 2022-05-26
person

valerii-derkach:
[DEMO LINK](https://valerii-derkach.github.io/react_person/)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -11,19 +11,19 @@ export const Person = ({ person }) => {
     partnerName,
   } = person;
 
+  const partner = sex === 'f' ? 'husband' : 'wife';
+
   return (
     <section className="Person">
       <h2 className="Person__name">{`My name is ${name}`}</h2>
       {age && (
         <p className="Person__age">{`I am ${age}`}</p>
       )}
-      {(isMarried && (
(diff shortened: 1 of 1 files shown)


volodya-leskiv (approved):
Good solution, Valerii. You'll be a Senior soon!

src/components/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+      {(isMarried && (
+        sex === 'f' ? (
+          <p className="Person__partner">{`My husband's name is ${partnerName}`}</p>
+        ) : (
+          <p className="Person__partner">{`My wife's name is ${partnerName}`}</p>
+        )
+      )) || 'I am not married'}
```
Anna-Tkachenko, line 26: Will be better to create separate variable before return statement in component, save here final married status message and use this variable in jsx

Anna-Tkachenko (changes requested):

src/components/Person.jsx
valerii-derkach: i hardly catch what did you want exactly, but i have fixed logic

Anna-Tkachenko (approved):
⭐️

PR #393, opened 2022-05-26
new commit fixed scss

volodya-leskiv:
[DEMO LINK](https://volodya-leskiv.github.io/react_person/)


src/App.js
```diff
@@ -28,22 +28,25 @@ const alex = {
...
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">My husband&apos;s name is Maksym</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person
```
alexandra-protyanova, line 31: you can make it as:
`<Person person={misha}`
and destruturate it in person

alexandra-protyanova (approved):

src/App.js
Anna-Tkachenko: Agree, will be better to pass full person instead of each separate person property

src/components/Person.jsx
```diff
@@ -0,0 +1,37 @@
...
+export const Person = ({
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+}) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${name}`}</h2>
+    {age && <p className="Person__age">{`I am ${age}`}</p>}
```
Anna-Tkachenko, line 14:
```suggestion
    {age && (
      <p className="Person__age">{`I am ${age}`}</p>
    )}
```

src/components/Person.jsx
```diff
@@ -0,0 +1,37 @@
...
+    <p className="Person__partner">
+      {isMarried
+        ? `My ${sex === 'f' ? 'husbands' : 'wife'} name is ${partnerName}`
```
Anna-Tkachenko, line 17: Will be better to create separate variable before return statement in component, save here final married status message and use this variable in jsx
Such way helps to avoid adding inner code logic in jsx for better understanding final html

Anna-Tkachenko (approved):

PR #395, opened 2022-05-27
Person cards

Yaroslav-Kononenko:
[DEMO LINK](https://Yaroslav-Kononenko.github.io/react_person/)


alexandra-protyanova (approved):

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const {
+    name, age, sex, isMarried, partnerName: pName,
+  } = person;
+
+  return (
+    // eslint-disable-next-line react/jsx-filename-extension
+    <section className="Person">
```
Anna-Tkachenko, line 11: Use small letters in class naming
```suggestion
    <section className="person">
```

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+Person.defaultProps = {
+  age: 0,
+};
+
+Person.propTypes = {
+  name: PropTypes.string.isRequired,
+  age: PropTypes.number,
+  sex: PropTypes.string.isRequired,
+  isMarried: PropTypes.bool.isRequired,
```
Anna-Tkachenko, line 37: What about partnerName?

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,38 @@
...
+    // eslint-disable-next-line react/jsx-filename-extension
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {`My ${sex === 'm' ? 'wife' : 'husband'} name is ${pName}`}
```
Anna-Tkachenko, line 19: Will be better to create separate variable before return statement with the result of married status message
And use this variable here to display some string

Anna-Tkachenko (changes requested):

PR #397, opened 2022-05-27
Solution

iwanttosmile4u:
[DEMO LINK](https://iwanttosmile4u.github.io/react_person/)


Danylo-Dziuban (approved):
Good job!👍

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,51 @@
...
+          ? `My ${partner}'s name is ${partnerName}`
+          : 'I am not married'}
+      </p>
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  person: {
+    name: '',
```
Anna-Tkachenko, line 37: You should add default props only for non required fields (age and partnerName)

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #400, opened 2022-05-27
First_commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_person/)


src/components/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        { name ? `My name is ${name}` : 'Unknown'}
```
alexandra-protyanova, line 16: it's not necessary, all people have names

alexandra-protyanova (approved):

src/App.js
```diff
@@ -28,22 +28,17 @@ const alex = {
...
 const App = () => (
   <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">My wife&apos;s name is Natasha</p>
-    </section>
+    <Person
+      person={misha}
+    />
```
Anna-Tkachenko, line 33:
```suggestion
    <Person person={misha} />
```

Anna-Tkachenko: Check point 2 in checklist
https://github.com/mate-academy/react_sum/blob/master/checklist.md

src/components/Person.jsx
Anna-Tkachenko: Agree, name is required field

src/components/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+      </h2>
+      <p className="Person__age">
+        { age && `I am ${age}`}
+      </p>
```
Anna-Tkachenko, line 20: You shouldn't render this tag if age is null
{age && (
  <p> ... </p>
)}

src/components/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+      <p className="Person__partner">
+        {(!isMarried && 'I am not married')
+        || (`My ${sex === 'm' ? 'wife' : 'husband'}'s name is ${partnerName}`)}
+      </p>
```
Anna-Tkachenko, line 24: This code looks too complicated
Create separate variables before return statement and store here marriage status
const personSex = sex === 'm' ? ... : ...;
const marriageStatus = `My ... name is ...`;

and use this variable in jsx

src/components/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+// eslint-disable-next-line react/no-typos
+Person.PropTypes = {
+  person: PropTypes.shape({
+    name: PropTypes.string,
+    age: PropTypes.number,
+    sex: PropTypes.string,
+    isMarried: PropTypes.bool,
+    partnerName: PropTypes.string,
```
Anna-Tkachenko, line 44: All fields should be required except the age and partnerName

Anna-Tkachenko (changes requested):

src/components/Person.jsx
igoree88: Fixed

src/components/Person.jsx
igoree88: Fixed

src/components/Person.jsx
igoree88: Maybe I didn't  understand what you mean, but this code works. It doesn't render this tag if age is null.

src/App.js
igoree88: Fixed

src/components/Person.jsx
igoree88: Fixed

Anna-Tkachenko (changes requested):

src/components/Person.jsx
igoree88: Thanks for explanation, I've understood. Fixed.

Denchepornyuk (approved):

PR #401, opened 2022-05-27
React Person Solution

59MrRobot:
- [DEMO LINK](https://59MrRobot.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import './Person.scss';
+import PropTypes from 'prop-types';
+
+export const Person = ({ personInfo }) => {
+  const {
+    name,
+    sex,
+    age,
+    isMarried,
+    partnerName,
+  } = personInfo;
+
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      <p className="Person__partner">
+        {isMarried
+          ? (`My ${partner}'s name is ${partnerName}`)
+          : ('I am not married')}
+      </p>
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  personInfo: {
+    age: 0,
+    partnerName: '',
+  },
+};
+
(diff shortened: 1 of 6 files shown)


src/components/Person/Person.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import './Person.scss';
+import PropTypes from 'prop-types';
```
LukaszGryziewicz, line 3: You should push the package.json (and maybe also lock?) after adding this to dependencies, it should fix the github tests

LukaszGryziewicz (approved):
Good job, check my comment about github tests

src/components/Person/Person.jsx
VinnieJ-2k20: package.json is enough 😅 
image

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,50 @@
...
+      </p>
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  personInfo: {
+    age: 0,
+    partnerName: '',
+  },
```
VinnieJ-2k20, line 39: This will work only if the whole person won't be passed as prop. Pass not the whole object, but separate props.

VinnieJ-2k20 (changes requested):
Almost done, need to fix some props.

.eslintrc.js
```diff
@@ -1,22 +1,14 @@
 module.exports = {
-  extends: ['@mate-academy/eslint-config-react', 'plugin:cypress/recommended'],
+  extends: ['@mate-academy/eslint-config-react-typescript', 'plugin:cypress/recommended'],
```
ihor-jpeg, line 2: but you are not using ts for this task

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,46 @@
...
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      <p className="Person__partner">
+        {isMarried
+          ? (`My ${partner}'s name is ${partnerName}`)
+          : ('I am not married')}
+      </p>
```
ihor-jpeg, line 30:
```suggestion
      <p className="Person__partner">
        {isMarried
          ? `My ${partner}'s name is ${partnerName}`
          : 'I am not married'
        }
      </p>
```

ihor-jpeg (approved):

PR #402, opened 2022-05-27
solution

LukaszGryziewicz:


src/Person/Person.jsx
```diff
@@ -0,0 +1,54 @@
...
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
```
efdalyalcin, line 12: you could directly destruct the object.

efdalyalcin (changes requested):
Great job! I just left a little comment

efdalyalcin (approved):
Well done, again sorry for your time !

src/Person/Person.jsx
```diff
@@ -0,0 +1,54 @@
...
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  person: {
+    age: 0,
+    partnerName: 'not provided',
+  },
+};
```
VinnieJ-2k20, line 44: This will work only if the whole person won't be passed as prop. Pass not the whole object, but separate properties.

VinnieJ-2k20 (changes requested):
Almost done, one thing left to fix.

ihor-jpeg (approved):

PR #403, opened 2022-05-27
solution

efdalyalcin:


src/components/Person/Person.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
```
LukaszGryziewicz, line 2: prop-types seems to not be pre-installed in this project, so you also should push the package.json to pass the tests on github

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      <p className="Person__age">
+        {age ? `I am ${age}` : null}
+      </p>
```
LukaszGryziewicz, line 22: I think that you should render the whole paragraph conditionally, otherwise you might end up with empty elements, like in the screenshot. 
image

LukaszGryziewicz (changes requested):
Good job, please check my comments 👍

LukaszGryziewicz (approved):
Nice 🥇

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {
+        age
+          ? (
+            <p className="Person__age">
```
VinnieJ-2k20, line 23: use `&&`

VinnieJ-2k20 (changes requested):
Almost done, one small thing to fix.

src/components/Person/Person.jsx
efdalyalcin: Actually I tried but then it prints `0` on the page for an unknown reason to me :)

src/App.jsx
```diff
@@ -0,0 +1,29 @@
...
+import Person from './components/Person/Person';
+
+const App = () => (
+  <div className="App">
+    <Person
+      name="Misha"
+      age={37}
+      sex="m"
+      isMarried
+      partnerName="Natasha"
```
ihor-jpeg, line 11: why not pass a single prop person?

src/components/Person/Person.jsx
ihor-jpeg: it prints zero because you have set 0 to be the default prop. do not make it default, just make it not required

ihor-jpeg (changes requested):

PR #405, opened 2022-05-28
Develop

AndriySimyachenko:
[DEMO LINK](https://AndriySimyachenko.github.io/react_person/)


alexandra-protyanova (approved):

src/Components/Person/Person.jsx
```diff
@@ -0,0 +1,53 @@
...
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? (`My ${sex === 'f' ? ('husband') : ('wife')}'s name is ${partnerName}`)
+          : ('I am not married')}
+      </p>
```
Anna-Tkachenko, line 32: Don't use ternary operator inside another ternary operator
Create separate variables before return statement which should save partner sex value and marriage status info
const partnerSex = sex === 'f' ? ... : ... ;
const marriageStatus = `My ... name is ...`;

and use this variable in jsx
Such approach make your code clearly

Anna-Tkachenko (changes requested):

PR #408, opened 2022-05-29
solution

Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -0,0 +1,51 @@
...
+    </>
+  );
+};
+
+Person.defaultProps = {
+  info: {
+    age: 0,
+    partnerName: 'Not Married',
+  },
+};
```
amitpatiljc, line 41: I think you should pass separate properties do not pass whole object.

amitpatiljc (changes requested):
Good job, Just one suggestion please check the comment.

VinnieJ-2k20 (changes requested):

ihor-jpeg (changes requested):
take a look at Paul's comments

src/components/Person/Person.jsx
```diff
@@ -1 +1,46 @@
...
+  age,
+  sex,
+  isMarried,
+  partnerName,
+}) {
+
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <>
```
meforoma, line 17: you have a `<section>` wrapper - you don't need a fragment in such case
you would need it if you have `1+` `top-level html tags` - to encapsulate them.
```suggestion
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,46 @@
...
+      <section className="Person">
+        <h2 className="Person__name">
+          {`My name is ${name}`}
+        </h2>
+        <p className="Person__age">
+          {age && (`I am ${age}`)}
+        </p>
```
meforoma, line 24: so, here you still generate the age element, but it has zero-value
Olya is zero-aged

you should not generate the whole element if there's no age
```suggestion
        {age && (
          <p className="Person__age">
            `I am ${age}`
          </p>)
        }
```

src/App.js
```diff
@@ -28,22 +29,9 @@ const alex = {
...
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person {...misha} />
+    <Person {...olya} />
+    <Person {...alex} />
```
meforoma, line 34: what's the idea of passing a copy of an object? (`{ ... }`)
pass an object as-is
```suggestion
    <Person {misha} />
    <Person {olya} />
    <Person {alex} />
```

meforoma (changes requested):

PR #410, opened 2022-05-29
Task complete

rHashTag:
[DEMO LINK](https://rHashTag.github.io/react_person/)


src/Component/Person/Person.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import './Person.scss';
+
```
amitpatiljc, line 3: I think we should use default proptype here as we have some properties absent for some people.

rHashTag: I check with Paul, it is optional on the task.

amitpatiljc: Right!

src/App.js
```diff
@@ -28,22 +29,11 @@ const alex = {
...
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">My wife&apos;s name is Natasha</p>
-    </section>
 
-    <section className="Person">
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">My husband&apos;s name is Maksym</p>
-    </section>
+    <Person props={misha} />
```
VinnieJ-2k20, line 33: Don't use `props` for prop name, it's very confusing.

VinnieJ-2k20 (approved):
Good job 👍 
Read the comment.

PR #411, opened 2022-05-30
Solution

huberciak:
[DEMO LINK](https://huberciak.github.io/react_person/)


andreisucman (approved):
Good job. Just one suggestion. I think you don't have to use interpolation inside a `<p>` tag.

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,60 @@
...
+      {
+        age
+          ? (
+            <p className="Person__age">
+              {`I am ${age}`}
+            </p>
+          )
+
+          : null
+      }
```
ihor-jpeg, line 32: use logical AND

src/components/Person/Person.jsx
```diff
@@ -0,0 +1,60 @@
...
+export const Person = ({
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+}) => {
+  function partner(s) {
+    return s === 'm' ? 'wife' : 'husband';
+  }
```
ihor-jpeg, line 14: i dont think you need a function for this. a simple variable and a ternary should be enough

src/components/App/App.jsx
```diff
@@ -0,0 +1,37 @@
...
+  age: 25,
+  sex: 'm',
+  isMarried: false,
+};
+
+export const App = () => (
+  <div className="App">
+    <Person {...misha} />
+    <Person {...olya} />
+    <Person {...alex} />
```
ihor-jpeg, line 33: it's not the best approach, but it's okay for this task

src/components/App/App.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Person } from '../Person';
+
+// eslint-disable-next-line
```
ihor-jpeg, line 4: it's not okay to use linter in order to then disable it

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #414, opened 2022-05-30
react_person task

Sardorbek1122:
[DEMO LINK](https://sardorbek1122.github.io/react_person/)
[React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline)

### DIFF
--- a/src/components/Person.jsx
+++ b/src/components/Person.jsx
@@ -39,17 +39,14 @@ export const Person = ({ person }) => {
 };
 
 Person.defaultProps = {
-  name: '',
   age: 0,
-  sex: 'm',
-  isMarried: false,
-  partnerName: 'not married',
+  partnerName: '',
 };
 
 Person.propTypes = {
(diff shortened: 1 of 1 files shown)


proznerka (approved):
I think we don't need to set defaults in this task, but rest of the code looks good to me.

src/components/Person.jsx
```diff
@@ -0,0 +1,55 @@
...
+  );
+};
+
+Person.defaultProps = {
+  name: '',
+  age: 0,
+  sex: 'm',
+  isMarried: false,
+  partnerName: 'not married',
+};
```
ihor-jpeg, line 47: it seems like the props that the component receives are an object. not individual properties

ihor-jpeg (changes requested):

meforoma (approved):
except for redundant `import './App.scss';`
it's ok

PR #415, opened 2022-05-30
task completed

andreisucman:


src/Person/Person.jsx
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
amitpatiljc, line 1: You didn't import react, Person.scss file and proptypes

amitpatiljc (changes requested):
Good job Mate! Please take a glance at comments.

src/Person/Person.jsx
andreisucman: true story...

src/App.jsx
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable react/jsx-filename-extension */
+import React from 'react';
+import './App.scss';
+import Person from './Person';
+
+// eslint-disable-next-line
```
ihor-jpeg, line 6: let's not disable linter

ihor-jpeg: you can create a json file with an array containing these

src/Person/Person.jsx
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable react/jsx-one-expression-per-line */
+/* eslint-disable react/jsx-filename-extension */
```
ihor-jpeg, line 2: you most likely need to disable these for the entire project. do that in .eslintrc file

src/Person/Person.jsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react';
+import './Person.scss';
+import PropTypes from 'prop-types';
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
ihor-jpeg, line 7: is you pass a single object as a prop (person), you would not need to disable linter here

src/Person/Person.jsx
```diff
@@ -0,0 +1,52 @@
...
+          I am {age}
+        </p>
+      )}
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  partnerName: null,
+};
```
ihor-jpeg, line 43: maybe this one is required? otherwise the paragraph with partner will not make much sense

ihor-jpeg: also, you dont have to set null as default

ihor-jpeg (changes requested):

src/Person/Person.jsx
andreisucman: Fixed.

src/App.jsx
```diff
@@ -0,0 +1,36 @@
...
+const alex = {
+  name: 'Alex',
+  age: 25,
+  sex: 'm',
+  isMarried: false,
+};
+
+const App = () => (
+  <div className="App">
+    <Person {...misha} />
```
VinnieJ-2k20, line 28: Don't spread props like this. Write them individually or pass the whole objedct.

VinnieJ-2k20 (changes requested):
One small thing left to fix.

ihor-jpeg (approved):

PR #452, opened 2022-07-02
Function Person

Andriy3323:
[DEMO LINK](https://andriy3323.github.io/react_person/)


sancho325 (approved):
gj

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? `My ${sex === 'm' ? 'wife' : 'husband'}'s name is ${partnerName}`
```
illiarerun, line 25: I'd move it to the variable, too much logic in JSX is bad

illiarerun (changes requested):

src/components/Person/Person.jsx
Andriy3323: there is only 1 ternary operator, I can't see reason to create variable. After all, this block of code is not reused

jstmpelowycz (approved):

PR #477, opened 2022-07-04
Solution

yanaost:
DEMO LINK](https://yanaost.github.io/react_person/)


KrisMur (changes requested):
Привіт все ок, але ти не додала Демолінку в рідмі файл

KrisMur (approved):

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
-// Write code here
+import React from 'react';
+import './Person.scss';
+
+export function Person({ person }) {
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${person.name}`}
```
illiarerun, line 8: Use desturctuirng to avoid writing `person.` everytime

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+      </h2>
+
+      <p className="Person__age">
+        {`I am ${person.age || null}`}
+      </p>
```
illiarerun, line 13: do you need a `p` with a `null` value? I don't think so. You can wrap your `p` tag with `age &&` condition

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+
+      <p className="Person__partner">
+        {person.isMarried ? `My hasbands name is ${person.partnerName}` : `I am not married`}
```
illiarerun, line 16: Let's consider that male person can have a wife, so you need to either render text with `husband's` or `wife's` word

illiarerun: Also, move it to the variable, too much going on here

illiarerun (changes requested):

src/components/Person/Person.jsx
yanaost: I prefer to destruct inside component

Olexandr6 (approved):

PR #507, opened 2022-07-18
first commit

pavel-gutsal:
https://pavel-gutsal.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -2,15 +2,15 @@
 import React from 'react';
 
 export const Person = ({ person }) => {
-  const { name, age, sex, isMarried, partnerName: pN } = person;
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
 // export function Person({ person }) {}
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName: pN } = person;
```
valerii-derkach, line 5: it is not ok to use `pN` as name of variable, use more understandable names
just use partnerName without shorts

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #517, opened 2022-07-28
added people

silchyn:
[DEMO](https://silchyn.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export function Person({ person }) {}
+import React from 'react';
+
+export function Person({ person }) {
+  return (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
-// export function Person({ person }) {}
+import React from 'react';
+
+export function Person({ person }) {
+  return (
+    <section className="Person">
```
VETALEK, line 5: let's destructure the `person`, so it'll be more convenient to use its props

VETALEK (changes requested):
GJ 👍 
commented

VETALEK (approved):
🔥

PR #543, opened 2022-08-01
Create checklist.md

solaryasha:


VinnieJ-2k20 (approved):
I would also add the case of creating separate variable instead of putting several cases to conditional rendering to avoid

```
{listEnabled && list.length && smthElse > 0 
  ? <ComponentOne />
  : <ComponentTwo />}
```

PR #557, opened 2022-08-02
solution

vitmixG:
[DEMO LINK](https://vitmixG.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,30 @@
-// export function Person({ person }) {}
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    isMarried,
+    partnerName,
+    sex,
+  } = person;
+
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,34 @@
-// export function Person({ person }) {}
+/* eslint-disable no-nested-ternary */
```
solaryasha, line 1: You don't need to disable this rule

src/components/Person/Person.jsx
```diff
@@ -1 +1,34 @@
...
+    age,
+    isMarried,
+    partnerName,
+    sex,
+  } = person;
+
+  let partner = '';
+
+  if (isMarried) {
+    if (sex === 'f') {
```
solaryasha, line 16: Avoid using nested `if` statements

solaryasha (approved):

PR #672, opened 2022-10-02
solution

Naveed-Shoukat:
[DEMO LINK](https://naveed-shoukat.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,33 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export function Person(person) {
+  const { name, age, sex, isMarried, partnerName } = person.person;
+  const partnerText = sex === 'm'
+    ? `${partnerName} is my wife`
+    : `${partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {
+        age && (
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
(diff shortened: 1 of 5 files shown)


fatihsamur (approved):
approved 👍

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export function Person(person) {
+  const { name, age, sex, isMarried, partnerName } = person.person;
```
ihor-jpeg, line 5: put each key on its own line

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export function Person(person) {
```
ihor-jpeg, line 4: it's only person, because you have one prop `person`. it's incorrect to call it person. but you can destructure and get the person prop

```suggestion
export function Person({ person }) {
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+        {`My name is ${name}`}
+      </h2>
+
+      {
+        age && (
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
+        )
+      }
```
ihor-jpeg, line 22:
```suggestion
      {age && (
        <p className="Person__age">
          {`I am ${age}`}
        </p>
      )}
```

fix it everywhere else

ihor-jpeg (changes requested):

PR #675, opened 2022-10-05
react_person

Hovera-Liubomyr:
https://Hovera-Liubomyr.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,24 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+      </h2>
+      <p className="Person__age">
+        {age
+          ? `I am ${age}`
+          : null}
+      </p>
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`
+          : `I am not married`}
```
Bohdan-Maksymiuk, line 20: better to calculate this values before return and then use variables here
do not overload jsx

PR #687, opened 2022-10-10
solution

vladimir-dublya:
[DEMO LINK](https://vladimir-dublya.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,29 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partner = sex === 'm'
+    ? 'wife'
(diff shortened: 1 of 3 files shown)


Anna-Borodaienko (approved):

naziamoff (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+      </h2>
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      {isMarried
+        ? `${partnerName} is my ${partner}`
+        : ('I am not married')
+      }
```
naziamoff, line 22: would be better to do it inside `<p>` element

naziamoff (approved):

src/components/Person/Person.jsx
naziamoff: I mean to create a new one

PR #688, opened 2022-10-10
solution was added

Maksym845:
- [DEMO LINK](https://Maksym845.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,29 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        && <p className="Person__age">{`I am ${age}`}</p>
+      }
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {sex === 'f'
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+      }
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {sex === 'f'
+              ? `${partnerName} is my husband`
+              : `${partnerName} is my wife`
+            }
+          </p>
+        )
```
Yeliseiev, line 20: It is not a good practice to use ternary operator in ternary operator, maybe you can calculate it from outside?

Yeliseiev (approved):
I will approve, but I am not sure you will pass mentors review

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+      {age
+        && <p className="Person__age">{`I am ${age}`}</p>
+      }
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {sex === 'f'
+              ? `${partnerName} is my husband`
+              : `${partnerName} is my wife`
+            }
```
Denchepornyuk, line 18: let's move this ternary operator in a variable and use the variable here

Denchepornyuk (changes requested):

Denchepornyuk (approved):

tetiana-ivanchenko (approved):

PR #690, opened 2022-10-10
react_person

DikanevYura:
https://DikanevYura.github.io/react_person/

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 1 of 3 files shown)


src/App.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
```
TarasHoliuk, line 3: if you create a separate folder `Person` with `Person.tsx`, better to create also `index.js` file with re-export:
```
export * from './Person.tsx;
```
then you'll be able to import from `index.js` file automatically, and import will look like
`import { Person } from './components/Person`;

TarasHoliuk (changes requested):
GJ! Added one comment

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+            ? 'husband'
+            : 'wife'
+          }`}
+        </p>
+      )
+      : (
+        <p className="Person__partner">
+          I am not married
+        </p>
+      )}
```
AlisaVlasova, line 33: давай позбавимось від вкладених тернарок, винеси щось у геттер

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #714, opened 2022-10-11
person

artem-musii:


src/components/Person/Person.jsx
```diff
@@ -1 +1,34 @@
...
+
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )
+      }
+      {isMarried
+        ? <p className="Person__partner">{`${partnerName} is my ${probablePartner}`}</p>
+        : <p className="Person__partner">I am not married</p>
```
Yevheniya-Deryka, line 30: seems to me it's better to use the ternary operator inside the paragraph not to duplicate it

Yevheniya-Deryka (approved):
GJ!
seems to me it's better to use the ternary operator inside the paragraph not to duplicate it

Andr1yK (approved):

PR #715, opened 2022-10-11
Solution

OleksandrGlulhovskiy:
[DEMO LINK](https://OleksandrGlulhovskiy.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,36 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partner = sex === 'm'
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? (`${partnerName} is my ${partner}`)
```
glugovsky, line 30: Doesn't seem to be necessary to use additional parentheses here
```suggestion
          ? `${partnerName} is my ${partner}`
```

glugovsky (approved):

src/components/Person/Person.jsx
naziamoff: I believe Serhii is right here

naziamoff (approved):
Good job! Just one small thing to fix

PR #721, opened 2022-10-12
React person

glugovsky:
[DEMO LINK](https://glugovsky.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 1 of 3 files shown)


OleksandrGlulhovskiy (commented):
GJ!

src/App.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
```
TarasHoliuk, line 3: if you create a separate folder `Person` with `Person.tsx`, better to create also `index.js` file with re-export:
```
export * from './Person.tsx;
```
then you'll be able to import from `index.js` file automatically, and import will look like
`import { Person } from './components/Person`;

TarasHoliuk (changes requested):
Nicely done! Added one comment

PR #734, opened 2022-10-13
feat: add person component

iryna-borniak:
[DEMO LINK](https://iryna-borniak.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,27 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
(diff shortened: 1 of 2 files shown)


tatyana22 (approved):
GJ!

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+        {`My name is ${name}`}
+      </h2>
+      {age
+        ? (
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
+        )
+        : null
+      }
```
naziamoff, line 24: think of a way to avoid ternary in this case, there's a simpler way. You don't need to return null

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #738, opened 2022-10-14
Solution

AniaHna:
[DEMO LINK](https://AniaHna.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,27 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const couple = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```
ArturNagaycev, line 4: I heard somewhere, that if we have more than 2-3 arguments, we must move each to the next line. 
But now it looks readable. 
I don't know :) 
Let a mentor will say

ArturNagaycev (approved):
🥇

src/components/Person/Person.jsx
naziamoff: agree that would be better to move each next to the new line. Doesn't seem to be critical in that case though

naziamoff (approved):
👍

PR #746, opened 2022-10-18
react Person

zhenija:
[DEMO LINK](https://zhenija.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,25 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+    {person.age && (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+    {person.age && (
+      <p className="Person__age">{`I am ${person.age}`}</p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.sex === 'f' ? (
+          `${person.partnerName} is my husband`
+        ) : (
+          `${person.partnerName} is my wife`
```
Bohdan-Maksymiuk, line 16: better to prepare needed string before return and use variable here

Bohdan-Maksymiuk (changes requested):
Good job, fix small comment

Bohdan-Maksymiuk (approved):

Bohdan-Maksymiuk (commented):
for readability better to pass into variable what function returned and then use variable
but it is also ok

PR #773, opened 2022-11-08
Solution

anastasia-babko:
[DEMO LINK](https://anastasia-babko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,22 +1,24 @@
 import React from 'react';
 
+function marriedInfo(info) {
+  return info === 'm' ? 'wife' : 'husband';
+}
+
 export const Person = ({ person }) => (
   <section className="Person">
     <h2 className="Person__name">
       {`My name is ${person.name}`}
     </h2>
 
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my`
+        + ` ${person.sex === 'm' ? 'wife' : 'husband'}`}
+      </p>
+    ) : 'I am not married'}
```
SanyaBratashchuk, line 20: 1. Split this logic and move part of it out of JSX.
2. `I am married` also should be in a `p` tag

SanyaBratashchuk (changes requested):

src/components/Person/Person.jsx
anastasia-babko: I'm not sure I correctly understood what you meant in the first one, but I kinda fixed it.

PR #788, opened 2022-11-15
slution

IOneStarI:
[DEMO LINK](https://ionestari.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,44 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partnerString = sex === 'm'
+    ? (<p className="Person__partner">{`${partnerName} is my wife`}</p>)
+    : ((<p className="Person__partner">{`${partnerName} is my husband`}</p>)
+    );
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : undefined
+      }
+      {isMarried
+        ? partnerString
+        : (<p className="Person__partner">I am not married</p>)
+      }
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  age: undefined,
+  partnerName: undefined,
+};
+
+Person.propTypes = {
+  name: PropTypes.number.isRequired,
+  age: PropTypes.number,
+  sex: PropTypes.string.isRequired,
+  isMarried: PropTypes.bool.isRequired,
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partnerString = sex === 'm'
+    ? (<p className="Person__partner">{`${partnerName} is my wife`}</p>)
+    : ((<p className="Person__partner">{`${partnerName} is my husband`}</p>)
+    );
```
Stepico, line 16: I think this is what we should call DRY, only one word changes, so maybe you can make a pattern, where you put all same things and just change wife or husband. Maybe it's worth to make an additional variable for this

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : undefined
+      }
+      {isMarried
+        ? partnerString
+        : (<p className="Person__partner">I am not married</p>)
```
Stepico, line 27: The same here, to manage this problem you repeated this for 3 times

Stepico (changes requested):
GJ, but I think you can come up with better solution, and I'm not sure but maybe it's worth adding styles related to the Person to the appropriate file

Stepico (approved):
Well done

src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
```
mykola-koval, line 3:
```suggestion
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : undefined
+      }
```
mykola-koval, line 25:
```suggestion
      {age &&
        <p className="Person__age">{`I am ${age}`}</p>
      }
```

mykola-koval (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,45 @@
...
+  const partner = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age &&
+        <p className="Person__age">{`I am ${age}`}</p>
+      }
```
mykola-koval, line 23: you probably shouldn't commit suggestions through github UI
this will give you an error in VSCode :)

mykola-koval (approved):

PR #790, opened 2022-11-15
Person component was added

iryna-lelyk:
[DEMO LINK](https://iryna-lelyk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,47 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+          ? `${partnerName} is my ${partner}`
+          : 'I am not married'
+        }
+      </p>
+    </section>
+  );
+};
+
+Person.propTypes = {
+  person: PropTypes.shape({
```
taraselnik, line 40: why do you use shape here? )))

taraselnik (approved):

src/components/Person/Person.jsx
iryna-lelyk: Because as argument in props we get object, that should have specific values (name, agge, sex, etc)

mykola-koval (approved):

PR #804, opened 2022-11-16
solution

MelnychukYana:
https://MelnychukYana.github.io/react_person/


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {!isMarried && <p className="Person__partner">I am not married</p>}
+      {isMarried && (
+        <p className="Person__partner">{`${partnerName} is my ${partner}`}</p>
```
Bohdan-Maksymiuk, line 13: looks very similar - so better prepare text before return in variable and use it here inside tag

Bohdan-Maksymiuk (changes requested):
try to fix comment

PR #805, opened 2022-11-16
v1

sergosdv1:
[DEMO LINK](https://DanilSerzh.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -9,15 +9,9 @@ export const Person = ({ person }) => {
     isMarried,
     partnerName,
   } = person;
-  let partner = '';
-
-  if (isMarried && sex === 'm') {
-    partner = 'wife';
-  } else if (isMarried && sex === 'f') {
-    partner = 'husband';
-  } else {
-    partner = null;
-  }
+  const partner = sex === 'm'
+    ? 'wife'
+    : 'husband';
 
   return (
     <section className="Person">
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,51 @@
...
+  } = person;
+  let partner = '';
+
+  if (isMarried && sex === 'm') {
+    partner = 'wife';
+  } else if (isMarried && sex === 'f') {
+    partner = 'husband';
+  } else {
+    partner = null;
+  }
```
Stepico, line 20: Use ternary here

src/components/Person/Person.jsx
```diff
@@ -1 +1,51 @@
...
+      {age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+      )}
+
+      <p className="Person__partner">
+        {partner === null
+          ? `I am not married`
+          : `${partnerName} is my ${partner}`}
```
Stepico, line 37: It's better to put this logic into an additional variable. Also, you already have variable isMarried, you don't need to take care of partner === null, your else in your if-else statement isn't necessity, just because you already can check everything you need with isMarried

Stepico (changes requested):
GJ, just some minor suggestions from me and maybe you should put all Person-related styles into Person.scss

src/components/Person/Person.jsx
```diff
@@ -1 +1,46 @@
...
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+  const partner = sex === 'm'
```
denys-danyliuk, line 12:
```suggestion
  } = person;
  
  const partner = sex === 'm'
```

denys-danyliuk (approved):

PR #808, opened 2022-11-16
person created

Esceype:
- [DEMO LINK](https://Esceype.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,29 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const personSex = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      <p className="Person__partner">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const personSex = sex === 'm'
```
Stepico, line 12: I guess it's a bad name, personSex is about male or female, but it's not obvious that you put there wife or husband

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      <p className="Person__partner">
+        {isMarried ? `${partnerName} is my ${personSex}` : `I am not married`}
```
Stepico, line 25: Each part of ternary should be placed on different lines and this logic is better to put into additional variable

Stepico (changes requested):
GJ, just few suggestions from me and we also have styles related to the person in the app.scss, maybe you should put them into person.scss

src/components/Person/Person.jsx
Esceype: In this task, this check was performed once and not repeated anywhere, so I don't think you need to create a separate variable for it

Stepico (approved):
Well done

Olexandr6 (approved):

PR #817, opened 2022-11-17
Solution

Bohdan-Kovalenko1992:
[DEMO LINK](https://Bohdan-Kovalenko1992.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerType = (sex === 'm') ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+      <p className="Person__age">
+        I am
+        {age}
+      </p>
+      )}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${partnerType}`
+          : 'I am not married'}
+      </p>
+    </section>
+  );
+};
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```
Stepico, line 4: when we have more than 3 variables it's more suitable to place each other on new line

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+  const partnerType = (sex === 'm') ? 'wife' : 'husband';
```
Stepico, line 5: in ternary you have to place each part of it on the new line, and this braces aren't necessity

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+      {age && (
+      <p className="Person__age">
+        I am
+        {age}
+      </p>
+      )}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${partnerType}`
+          : 'I am not married'}
```
Stepico, line 19: This logic is better to put into additional variable

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerType = (sex === 'm') ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+      <p className="Person__age">
+        I am
+        {age}
```
Stepico, line 13: I guess this is the reason your tests were failed, there is no space between them, use the syntax from the string above, where you print person's name

Stepico (changes requested):
GJ, just few suggestions from me, also we have some styles related to the person that are in App.scss, maybe you should put them in person.scss

PR #818, opened 2022-11-17
develop

Ostapchuk-Y-O:
[DEMO LINK](https://Ostapchuk-Y-O.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,35 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partner = (sex === 'm')
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <>
+      <section className="Person">
+        <h2 className="Person__name">
+          {`My name is ${name}`}
+        </h2>
+
+        {age && (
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+          <p className="Person__age">{`I am ${age}`}</p>
+        )}
+
+        <p className="Person__partner">
+          {(isMarried)
+            ? `${partnerName} is my ${partner}` : `I am not married`}
+        </p>
+
+      </section>
+    </>
```
Stepico, line 33: You don't need fragment here, you already have only one parent element

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+          {`My name is ${name}`}
+        </h2>
+
+        {age && (
+          <p className="Person__age">{`I am ${age}`}</p>
+        )}
+
+        <p className="Person__partner">
+          {(isMarried)
+            ? `${partnerName} is my ${partner}` : `I am not married`}
```
Stepico, line 29: Each part of the ternary operator should be placed at the beginning of a new line and maybe you can put this logic in additional variable

Stepico (changes requested):
GJ. just some minor suggestions from me, and we have some person's styles in the App.scss, maybe you should to put them into Person.scss

Stepico:
Also I guess you should add propTypes to your solution

denys-danyliuk (approved):

PR #820, opened 2022-11-17
React person

referee1988:
[DEMO LINK](https://referee1988.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,47 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partnerType = (sex === 'm')
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age && (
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+  );
+};
+
+Person.propTypes = {
+  person: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+    age: PropTypes.number,
+    sex: PropTypes.string.isRequired,
+    isMarried: PropTypes.bool.isRequired,
+    partnerName: PropTypes.string.isRequired,
```
Stepico, line 45: partnerName isn't required, have a look at Alex

src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {(isMarried)
+          ? `${partnerName} is my ${partnerType}`
+          : `I am not married`
+        }
```
Stepico, line 33: It's better to put this logic into an additional variable

Stepico (changes requested):
GJ, just some minor suggestions from me and maybe you should put Person-related styles to Person.scss, because now they are in App.scss

src/components/Person/Person.jsx
referee1988: I disagree.
This logic is very simple and is not repeated anywhere else.

denys-danyliuk (approved):

PR #834, opened 2022-11-29
Solution

vitaliidox:
https://vitaliidox.github.io/react_person/


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+    }
+    <p className="Person__partner">
+      {(() => {
+        if (person.isMarried) {
+          return person.sex === 'm'
+            ? `${person.partnerName} is my wife`
+            : `${person.partnerName} is my husband`;
+        }
+
+        return `I am not married`;
```
vlad-demchuk, line 18: you shouldn't use a function here. just use conditional rendering with logical `&&`
https://reactjs.org/docs/conditional-rendering.html

vlad-demchuk: and use destructuring (checklist)
image
image

vlad-demchuk (changes requested):

tetiana-ivanchenko (approved):

PR #865, opened 2022-12-22
solution

Sergii-Osedchuk:
[DEMO LINK](https://Sergii-Osedchuk.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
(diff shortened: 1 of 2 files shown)


4ebupel (approved):
Not confident about nested ternary but if test are working, it's okay, I guess )

Sergii-Osedchuk:
> Not confident about nested ternary but if test are working, it's okay, I guess )

Me too) But if nested 'for loop' is ok, why ternary don't. Anyway we will check it out soon)

solaryasha (approved):
👍

PR #883, opened 2022-12-23
react person

victoria-kovalenko:
[DEMO LINK](https://victoria-kovalenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,32 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = (props) => {
+  const {
+    person,
+  } = props;
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : null
+      }
+
+      {isMarried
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = (props) => {
+  const {
+    person,
+  } = props;
```
Sveta-Kryukova, line 6: you can skip this part and write directly person insted of props)

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : null
+      }
```
Sveta-Kryukova, line 24: Good, but it shouldn't be rendered if there as no age. So try to use {age && (<...

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+
+      {isMarried
+        ? (<p className="Person__partner">{`${partnerName} is my ${sex === 'm' ? 'wife' : 'husband'}`}</p>)
+        : (<p className="Person__partner">I am not married</p>)
+      }
+    </section>
```
Sveta-Kryukova, line 30: Good, but for better understanding use a constant with {sex === 'm' ? 'wife' : 'husband'}

Sveta-Kryukova (changes requested):

valerii-derkach (approved):

PR #890, opened 2022-12-24
solution

kenvastolord:
[DEMO LINK](https://kenvastolord.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,26 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  const couple = (sex === 'm')
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {age && (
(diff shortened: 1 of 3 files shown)


RomeoKs (commented):
Looks good!

With only a few minor remarks/improvements.
Even though it was not stated in the task description, I would do the following:

1. move the styles for the Person from App.scss => Person.scss
2.  I would create a index.jsx in the Person components folder and export with `export * from './Person';`

Because when you will build larger apps, it will be easier for you to maintain the files/components/styles if you organize them in folders.

This task doesn't necessary need what I've written above and I'm sure that you'll consider my comment in the future.

Keep it up Kenneth!

RomeoKs (approved):

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  const couple = (sex === 'm')
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
```
naziamoff, line 12: better to move text to the new line, it will look much more readable
```suggestion
      <h2 className="Person__name">
        {`My name is ${name}`}
      </h2>
```

naziamoff (approved):

PR #897, opened 2022-12-25
Develop

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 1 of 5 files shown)


src/App.jsx
```diff
@@ -25,21 +26,18 @@ export const alex = {
...
+    />
+
+    <Person
+      person={olya}
+    />
+
+    <Person
+      person={alex}
+    />
+
```
dolnaro, line 41: I think you don't need to separate the Person function component line when you have only a single property. But of course, it is a preference :)

```
<Person person={misha} />
<Person person={olya} />
<Person person={alex} />
```

dolnaro (approved):

dolnaro:
All looks good.

src/App.jsx
naziamoff: agree with @dolnaro here

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partner = sex === 'f' ? 'husband' : 'wife';
```
naziamoff, line 6: ternaries should be multilined

naziamoff (approved):

PR #907, opened 2022-12-26
react_person

Mohameddahi:
[DEMO LINK](https://mohameddahi.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,26 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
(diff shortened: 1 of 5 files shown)


dolnaro (approved):
Look good to me.

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export const Person = ({ person }) => (
```
naziamoff, line 4: let's destructure person here

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${person.sex === 'm'
+          ? ('wife')
+          : ('husband')}`}
```
naziamoff, line 20: would be better to separate this calculation and render conditionally inside only one `<p>` tag

naziamoff (approved):

PR #917, opened 2022-12-27
React Person

YanaEvaGonchar:
- [DEMO LINK](https://YanaEvaGonchar.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,36 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const isNotMarried = isMarried
+    ? ''
+    : 'I am not married';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const isNotMarried = isMarried
```
romanblyashuk94, line 12: when variables name starts from "is...", it means that values have been boolean type.

isNotMarried  = empty string  - it looks a little bit strange and not obvious.

You can name it  like "notMarriedStatusMessage"

But it it will be better to use next logic:

```suggestion
  const partnerType = (sex === 'm')
    ? `wife`
    : `husband`;
    
    and then, in murkup....
    
    <p className="Person__partner">
        {
          isMarried
            ? `${partnerName} is my ${partnerType}`
            : `I am not married`
        }
      </p>
```

romanblyashuk94 (changes requested):

src/components/Person/Person.jsx
YanaEvaGonchar: fixed!

romanblyashuk94 (approved):
👍

solaryasha (approved):
👍

PR #919, opened 2022-12-27
Add solution

VladimirEdelmann:
# Task
Implement a `Person` component rendering a person details using the given markup
and use it 3 times inside the `App` instead of static markup.

- if a person is not marriend - print `I am not married` message.

# Implementation
[DEMO LINK](https://vladimiredelmann.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,41 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const hasPartner = () => {
+    if (!isMarried) {
+      return 'I am not married';
+    }
+
+    return sex === 'm'
+      ? `${partnerName} is my wife`
+      : `${partnerName} is my husband`;
+  };
+
+  const renderAge = () => (
+    <p className="Person__age">
+      {`I am ${age}`}
+    </p>
+  );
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,41 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {person.age && renderAge()}
+
+      <p className="Person__partner">
+        {hasPartner()}
```
romanblyashuk94, line 37: nice usage of helpers function for conditional rendering of maried status but name of function can be more declarating. Like "getMariedStatusMessage" or something like that)

src/components/Person/Person.jsx
```diff
@@ -1 +1,41 @@
...
+    </p>
+  );
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {person.age && renderAge()}
```
romanblyashuk94, line 34: I think it's redundant.

its better, when murkup placed with murkup)

and you can get age from alredy destructured person object.
Just "age" - not "person.age".

```suggestion
	{age && <p className="Person__age">{`I am ${age}`}</p>}
```

romanblyashuk94 (changes requested):

romanblyashuk94 (approved):
Good job!

valerii-derkach (approved):

PR #946, opened 2023-01-18
Task was done

Sergey-Mironenko:


Sergey-Mironenko:
https://Sergey-Mironenko.github.io/react_person/

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const iff = (condition, then, otherwise) => (
+    condition ? then : otherwise
+  );
```
vlad-demchuk, line 6: it will be the same if you write this ternary in jsx, and even clearer

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {' '}
+        {person.name}
```
vlad-demchuk, line 13: use destructuring for `person`

vlad-demchuk (changes requested):

AlisaVlasova (approved):

PR #951, opened 2023-01-19
solution

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 export const Person = ({ person }) => {
   const {
@@ -8,6 +9,7 @@ export const Person = ({ person }) => {
     isMarried,
     partnerName,
   } = person;
+
   const partner = sex === 'm'
     ? 'wife'
     : 'husband';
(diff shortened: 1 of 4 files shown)


GarryRocksman (approved):
GJ 🔥 just i think you need to add prop-types & default props

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
```
Denchepornyuk, line 10: add PropTypes for this component

denys-cheporniuk (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,51 @@
...
+  partnerName: null,
+};
+
+Person.propTypes = {
+  name: PropTypes.string.isRequired,
+  age: PropTypes.number,
+  sex: PropTypes.string.isRequired,
+  isMarried: PropTypes.bool.isRequired,
+  partnerName: PropTypes.string,
+};
```
Denchepornyuk, line 51: does this propsTypes work?
you don't get all these fields as props, you get only person with all these fields inside

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #960, opened 2023-01-21
solution

AnnyKoval:
[DEMO LINK](https://AnnyKoval.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,30 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
```
mykhalenych, line 3:
```suggestion
export const Person = ({ name, age, sex, isMarried, partnerName }) => {
```

mykhalenych (changes requested):

src/components/Person/Person.jsx
AnnyKoval: Done

Virass (approved):

PR #966, opened 2023-01-22
React Person

nkondd:
[DEMO LINK](https://nkondd.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,50 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,50 @@
...
+        }
+      </p>
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  age: null,
+  partnerName: null,
+};
```
Denchepornyuk, line 42: here should be person with `and ` and `partnerName` properties

Denchepornyuk (changes requested):

denys-cheporniuk (approved):

PR #967, opened 2023-01-22
Person component implemented

oleh-holovnykh:
[DEMO LINK](https://oleh-holovnykh.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,53 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+  const partnerStatus = (sex === 'm')
+    ? 'wife'
+    : 'husband';
+
+  return (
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,53 @@
...
+    partnerName: PropTypes.string,
+  }),
+};
+
+Person.defaultProps = {
+  person: {
+    name: 'Jeanne Louise Calment',
+    age: 122,
+    sex: 'f',
+    isMarried: true,
```
andrii-andriushchenko, line 50: I'm not sure, will it works correct, cause `age` and `isMarried` can be omitted , but you set default value for it

andrii-andriushchenko (approved):
looks good 🚀 
just check my comment

src/components/Person/Person.jsx
oleh-holovnykh: 🤔 thanks, it is interesting
i checked this case
the default value will only be replaced by defaultProps if no object is passed. If an object is passed without a property, the property will be undefined.
It is possible to set a default value for a property during destructuring. I think this is not necessary in this task, but this is not accurate)

alexandra-protyanova (approved):
GJ 🌟

PR #970, opened 2023-01-23
Develop

OlehSharha:
[DEMO LINK](https://OlehSharha.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,48 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age = null,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partnerSex = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+      </p>
+    </section>
+
+  );
+};
+
+Person.defaultProp = {
+  name: null,
+  partnerName: null,
+};
```
Anastasiia-Khudych, line 39: There is no need to set defaultProp for .isRequired elements. You should write here the default values for `partnerName` and `age` (because they could be not passed)  and not write `isRequired` for these values in Person.propTypes.

Anastasiia-Khudych (changes requested):
Good job!

src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+
+Person.defaultProps = {
+  partnerName: null,
+  age: null,
+};
+
+Person.propTypes = {
+  name: PropTypes.string.isRequired,
+  age: PropTypes.number,
+  sex: PropTypes.string.isRequired,
```
Denchepornyuk, line 44: use oneOf from PropTypes for `sex` instead of string

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #971, opened 2023-01-23
Person component was created

Coyoote:
[DEMO LINK](https://Coyoote.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -21,15 +21,16 @@ export const Person = ({ person }) => {
       </h2>
 
       {age && (
-        <p className="Person__age">
-          {`I am ${age}`}
-        </p>
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
       )}
 
+      {!isMarried && (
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,54 @@
...
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+      )}
+
+      {!isMarried && (
+        <p className="Person__partner">I am not married</p>)}
+
+      {isMarried && (
+        <p className="Person__partner">{`${partnerName} is my ${partnerSex}`}</p>
```
GarryRocksman, line 33: you can do it with ternary to avoid duplication

GarryRocksman (approved):
GJ 🔥 but i think needs some fixes

src/components/Person/Person.jsx
Coyoote: Agreed, fixed it right away

alexandra-protyanova (approved):

PR #973, opened 2023-01-23
Solution

maasiimka:


src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```
vanst1, line 5:
```suggestion
  const { 
    name, 
    age, 
    sex, 
    isMarried, 
    partnerName
  } = person;
```

vanst1 (approved):
Good job:)

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+        {isMarried
+          ? partnerMessage
+          : 'I am not married'
+        }
+      </p>
+    </section>
+  );
+};
+
+Person.propTypes = {
```
Denchepornyuk, line 24: add default types for component

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      <p className="Person__partner">
+        {isMarried
+          ? partnerMessage
+          : 'I am not married'
+        }
+      </p>
+    </section>
```
Denchepornyuk, line 20: add some space between elements

Denchepornyuk (changes requested):

src/components/Person/Person.jsx
maasiimka: There is render by condition for not required props. Are default props needed? Or you mean 'types' like in TS? Maybe I don't understand.

Denchepornyuk: For some fields from the object, we can add default props, if they are not passed, then the default values will be used

Denchepornyuk (approved):

PR #982, opened 2023-01-23
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,24 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  let partner = (<p className="Person__partner">I am not married</p>);
+
+  if (isMarried && sex === 'f') {
+    partner = (<p className="Person__partner">{`${partnerName} is my wife`}</p>);
+  }
+
+  if (isMarried && sex === 'm') {
+    partner = (<p className="Person__partner">{`${partnerName} is my husband`}</p>);
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : null}
+      {partner}
+    </section>
+  );
+};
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+  if (isMarried && sex === 'm') {
+    partner = (<p className="Person__partner">{`${partnerName} is my wife`}</p>);
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        ? (<p className="Person__age">{`I am ${age}`}</p>)
+        : null}
```
AlexPro100pro, line 20: may be better use {age && ...} if you add this messege only when age is defined

src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+  const { name, age, sex, isMarried, partnerName } = person;
+  let partner = (<p className="Person__partner">I am not married</p>);
+
+  if (isMarried && sex === 'f') {
+    partner = (<p className="Person__partner">{`${partnerName} is my husband`}</p>);
+  }
+
+  if (isMarried && sex === 'm') {
+    partner = (<p className="Person__partner">{`${partnerName} is my wife`}</p>);
+  }
```
AlexPro100pro, line 13: use Conditional (ternary) operator

Denchepornyuk: agree, do it with ternary operator
You don't need condition for all tag `p`, you need it only for text inside this tag

src/components/Person/Person.jsx
Denchepornyuk: agree && looks better in this case

src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
```
Denchepornyuk, line 3: add PropTypes and default types for this component

Denchepornyuk (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+        ? (<p className="Person__partner">{`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`}</p>)
+        : (<p className="Person__partner">I am not married</p>)}
+    </section>
+  );
+};
+
+Person.propTypes = {
+  name: propTypes.string.isRequired,
+  age: propTypes.number,
+  sex: propTypes.string.isRequired,
```
Denchepornyuk, line 21: use `oneOf` for sex instead of `string`

Denchepornyuk (approved):

PR #984, opened 2023-01-23
commit

yaroslavna-horban:


src/components/Person/Person.jsx
```diff
@@ -1 +1,52 @@
...
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const marriedWoman = isMarried && sex === 'f';
+  const marriedMan = isMarried && sex === 'm';
```
GarryRocksman, line 14: you can do it in one line with ternary

src/components/Person/Person.jsx
```diff
@@ -1 +1,52 @@
...
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {!isMarried && 'I am not married'}
+        {marriedMan && `${partnerName} is my wife`}
+        {marriedWoman && `${partnerName} is my husband`}
```
GarryRocksman, line 31: here you also can do it easier with ternary

GarryRocksman (changes requested):
GJ 🔥 but i think needs some fixes

PR #1038, opened 2023-02-16
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,31 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {' '}
+      {person.name}
+    </h2>
+    {person.age && (
+      <p className="Person__age">
+        I am
+        {' '}
+        {person.age}
+      </p>
+    )}
+    {
+      person.isMarried ? (
+        <p className="Person__partner">
+          {person.partnerName}
+          {' '}
+          is my
+          {' '}
+          {person.sex === 'm' ? 'wife' : 'husband'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )
+    }
+  </section>
+);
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {' '}
+      {person.name}
```
witflash, line 8: Try to do it with string interpolation

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+    </h2>
+    {person.age && (
+      <p className="Person__age">
+        I am
+        {' '}
+        {person.age}
```
witflash, line 14: Here and below too

witflash (changes requested):
Check my comment and try to fix it

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+      person.isMarried ? (
+        <p className="Person__partner">
+          {person.partnerName}
+          {' is my '}
+          {person.sex === 'm' ? 'wife' : 'husband'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )
+    }
```
nazarmatsevych, line 23: Try not to use nested ternary oparator, you can move this part person.sex === 'm' ? 'wife' : 'husband' to separate variable for example pertner and use this variable here.
Also you don't need to create two `<p>` tags, you can create one and change only content using ternary

nazarmatsevych (changes requested):
Almost done,
pls check my comments

src/components/Person/Person.jsx
vlad-trundenko: thanks)

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
```
polosanya, line 3: You can destructure the person to get all the needed props of it
https://medium.com/@lcriswell/destructuring-props-in-react-b1c295005ce0

polosanya (approved):
Good job!

PR #1064, opened 2023-03-02
result

novak-k:
[DEMO LINK](https://novak-k.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const checkPartner = person.sex === 'm'
+    ? `${person.partnerName} is my wife`
+    : `${person.partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      { person.age ? (<p className="Person__age">{`I am ${person.age}`}</p>) : null }
+      <p className="Person__partner">
+        { person.isMarried ? checkPartner : 'I am not married' }
+      </p>
+    </section>
+  );
+};
--- a/README.md
+++ b/README.md
@@ -11,4 +11,4 @@ and use it 3 times inside the `App` instead of static markup.
 ## Instructions
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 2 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const checkPartner = person.sex === 'm'
```
sTorba24, line 4:
```suggestion
export const Person = ({ person }) => {
  const { sex, partnerName, name, age, isMarried } = person
  const checkPartner = person.sex === 'm'
```

sTorba24: you can destructure an object to not repeat using `person.{field name}` many times :)

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+    ? `${person.partnerName} is my wife`
+    : `${person.partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      { person.age ? (<p className="Person__age">{`I am ${person.age}`}</p>) : null }
```
sTorba24, line 11:
```suggestion
      { person.age && (
        <p className="Person__age">{`I am ${person.age}`}</p>
       )}
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const checkPartner = person.sex === 'm'
+    ? `${person.partnerName} is my wife`
+    : `${person.partnerName} is my husband`;
```
sTorba24, line 6:
```suggestion
  const personPartner = person.sex === 'm'
    ? `${person.partnerName} is my wife`
    : `${person.partnerName} is my husband`;
```

sTorba24: checkPartner sounds like function but you have variable :)
Also as you can see difference is only in partner naming `wife/husband` so you can keep only this part in variable and other text you can keep in your jsx

sTorba24 (changes requested):
Good work! Lets make it better :)

src/components/Person/Person.jsx
novak-k: + sure

src/components/Person/Person.jsx
novak-k: + done

src/components/Person/Person.jsx
novak-k: done, as much as I understand it!

mykhalenych (approved):

PR #1087, opened 2023-03-10
person component was implement

KuharchukDmytro:
https://KuharchukDmytro.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,53 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Person.scss';
+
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,53 @@
...
+        )
+        : (
+          <p className="Person__partner">I am not married</p>
+        )}
+    </section>
+  );
+};
+
+Person.defaultProps = {
+  age: null,
```
MarkBabich, line 40: Useless as for me. If `age` is not passed, then it will be in the conditions give `false`. Rendering `Person__age` won't happen in that case.

MarkBabich (approved):
GJ :)

OleksiiBlezniuk (approved):

PR #1124, opened 2023-03-13
Develop

kuzmi491:


src/components/Person/Person.jsx
```diff
@@ -1 +1,43 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+function findPartner(partnerName, sex, isMarried) {
```
Andr1yK, line 4: use a better name, you are not using this function to find someone

src/components/Person/Person.jsx
```diff
@@ -1 +1,43 @@
...
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && (
+        <>
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
+        </>
```
Andr1yK, line 35: if age = 0 I will see 0 :)

Andr1yK (changes requested):

PR #1142, opened 2023-03-20
Develop

anna-tovkailo:
https://anna-tovkailo.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,24 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person: {
+  name, age, sex, isMarried, partnerName,
+} }) => {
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {` ${name}` }
+      </h2>
+      {age && (
+        <p className="Person__age">
+          I am
+          {' '}
+          {age}
```
OleziO, line 18: Make a template string from this.

[More info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO (changes requested):

mykhalenych (approved):

PR #1145, opened 2023-03-26
all done

NazikVr:
[DEMO LINK](https://nazikvr.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -9,28 +9,24 @@ export const Person = ({ person: {
   sex,
   isMarried,
   partnerName,
-} }) => {
-  const gender = sex;
+} }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is {name}
+    </h2>
 
-  return (
-    <section className="Person">
-      <h2 className="Person__name">
-        My name is {name}
-      </h2>
+    {age && (
+    <p className="Person__age">
+      I am {age}
+    </p>
+    )}
 
-      {age && (
-        <p className="Person__age">
-          I am {age}
-        </p>
-      )}
-
-      {sex && isMarried ? (
-        <p className="Person__partner">
-          {partnerName} is my {gender === 'm' ? 'wife' : 'husband'}
-        </p>
-      ) : (
-        <p className="Person__partner">I am not married</p>
-      )}
-    </section>
-  );
-};
+    {sex && isMarried ? (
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
-// export const Person = ({ person }) => ();
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
+// "react/jsx-one-expression-per-line": [<enabled>, { "allow": "none"|"literal"|"single-child" }];
```
OleziO, line 3: Remove all comments from the code.

OleziO (changes requested):

polosanya (changes requested):
Almost done!

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+        My name is {name}
+      </h2>
+
+      {age && (
+        <p className="Person__age">
+          I am {age}
+        </p>
+      )}
+
+      {sex && isMarried ? (
```
lerastarynets, line 27: The 'sex' check in the condition is not necessary since 'sex' is always expected to be a string (either 'm' or 'f'). Only checking 'isMarried' should suffice.

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+// eslint-disable-next-line react/jsx-one-expression-per-line
+
+export const Person = ({ person: {
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+} }) => {
+  const gender = sex;
```
lerastarynets, line 13: The 'gender' variable is redundant, as it is an exact copy of the 'sex' variable. You can use the 'sex' variable directly in your code.

src/components/Person/Person.jsx
lerastarynets: Still not fixed)

lerastarynets (changes requested):
Left few comments on how to improve your code) Pls fix all the previous comments
Pls ask for help in **fe_chat0** if you don't understand something

src/components/Person/Person.jsx
OleziO: Still not fixed

OleziO (changes requested):

lerastarynets (approved):

PR #1164, opened 2023-04-03
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -3,31 +3,34 @@ import PropTypes from 'prop-types';
 
 export const Person = (
   { person },
-) => (
-  <section className="Person">
-    <h2 className="Person__name">
-      {`My name is ${person.name}`}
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,40 @@
...
+    <p className="Person__partner">
+      {person.isMarried
+        ? (
+          `${person.partnerName} is my `
+            + `${person.sex === 'f' ? 'husband' : 'wife'}`
+        )
+        : (
+          'I am not married'
+        )
+        }
```
Volodymyr-Melnyk-Fron10der, line 27: generate this string out of jsx, create some function or just spit it into a few variables with ternary operators

Volodymyr-Melnyk-Fron10der (changes requested):

qoa11a (approved):

PR #1178, opened 2023-04-08
Person component has been implemented

bojkovladislav:


src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+        {age && `I am ${age}`}
+      </p>
+
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my
+          ${sex === 'm'
+            ? 'wife'
+            : 'husband'}`
+          : 'I am not married'}
```
Volodymyr-Melnyk-Fron10der, line 22: move this logic out of jsx, it's too complicated and hard to read. do not afraif even create 2 variables such status, marriedString

Volodymyr-Melnyk-Fron10der (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      <p className="Person__age">
+        {age && `I am ${age}`}
+      </p>
```
DarkMistyRoom, line 15: apply conditional rendering to the whole `p`

DarkMistyRoom: that's why tests failed

DarkMistyRoom (approved):

PR #1179, opened 2023-04-08
function react_person works

BondarenkoDmitriy:


src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, ifMarrid = false, partnerName } = person;
+
+  const partnerSex = sex === 'm'
+    ? 'wife'
+    : 'husbend';
+
+  const personMarrid = ifMarrid
```
makandriival, line 10: in your demo i can see that all the people are "not married". You should take a look at the logic inside this variable, may be some variable names are not right.

makandriival (changes requested):
The deployed demo is not working properly, the reason might be in the names of variables "ifMarrid" in particular. Other than that GJ!🤞🏽

src/components/Person/Person.jsx
```diff
@@ -1 +1,41 @@
...
+    ? `${partnerName} is my ${partnerSex}`
+    : 'I am not married';
+
+  const personAge = age
+    ? (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+    )
+    : null;
```
SerhiiKirik, line 26:
```suggestion
  const personAge = age && (
    p className="Person__age">
      {`I am ${age}`}
    </p>
  )
```

SerhiiKirik (commented):
Great 🔥

PR #1190, opened 2023-04-10
Solution

Zad0rnik:
[DEMO LINK](https://Zad0rnik.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,38 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  let familyStatus = 'wife';
+
+  if (sex === 'f') {
+    familyStatus = 'husband';
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,38 @@
...
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  let familyStatus = 'wife';
+
+  if (sex === 'f') {
+    familyStatus = 'husband';
+  }
```
yuliia-anisimova, line 16: I think you can combine these lines using the ternary operator and get rid of 'if'

src/components/Person/Person.jsx
```diff
@@ -1 +1,38 @@
...
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        { isMarried
+          ? (`${partnerName} is my ${familyStatus}`)
+          : 'I am not married'
```
yuliia-anisimova, line 33: I think it would be better if you move this to the variable

yuliia-anisimova (changes requested):
Awesome🔥But I think you should fix a couple of things

src/components/Person/Person.jsx
SerhiiKirik: agree, it would be better

SerhiiKirik (approved):
Well done 🚀

PR #1228, opened 2023-04-19
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,8 +1,12 @@
 /* eslint-disable react/jsx-one-expression-per-line */
 import React from 'react';
 
-export const Person = ({ person }) => {
-  const { name, age, isMarried, sex, partnerName } = person;
+export const Person = ({
+  person: { name, age, isMarried, sex, partnerName },
+}) => {
+  const personPartnerText = isMarried
+    ? `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`
+    : 'I am not married';
 
   return (
     <section className="Person">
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
-// export const Person = ({ person }) => ();
+/* eslint-disable react/jsx-one-expression-per-line */
+import React from 'react';
+
+export const Person = ({ person }) => {
```
Radoslaw-Czerniawski, line 4: You could destructure instead of creating a variable in the body of the component
```
export const Person = ({ person: { name, age, // etc} }) => {
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+
+      {age && <p className="Person__age">I am {age}</p>}
+
+      <p className="Person__partner">
+        {
+          isMarried
```
Radoslaw-Czerniawski, line 15: You could calculate this and assign to const in the body of the component,
```
const personPartnerText = isMarried  ? `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'` : 'I am not married'
```
then just display here:
```
{personPartnerText}
```

Radoslaw-Czerniawski (changes requested):
Good pr, correct the destructuring of the props, with calculating and assigning to the value is optional but i personally like to separate the logic as much as possible from the presentational part

DorotaLeniecDev (approved):
GJ!

PR #1229, opened 2023-04-20
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${person.name}`}
+      </h2>
+      <p className="Person__age">
+        {person.age
```
mykhalenych, line 10: Instead of rendering 'null', you can just omit the entire 'Person__age' paragraph when there is no age provided. Use conditional rendering with '&&' operator 👍

mykhalenych (approved):

PR #1230, opened 2023-04-20
Solution

8Melya8:
[DEMO LINK](https://8melya8.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
```
Viktor-Kost, line 4: it's a good practice to use Object Destructuring
```suggestion
export const Person = ({ person }) => (
  const { name, age, isMarried, partnerName, sex } = person;
  
  return (
    <section className="Person">
      {{your__code}}
    </section>
  )
)
```

Viktor-Kost (approved):
Good job!

Add one recommendation for you

PR #1242, opened 2023-04-26
solution

kq333:
[DEMO LINK](https://kq333.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,7 @@
+/* eslint-disable import/named */
 import React from 'react';
 import './App.scss';
+import Person from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
@@ -24,22 +26,10 @@ export const alex = {
 };
 
 export const App = () => (
-  <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
+  <>
+    <Person {...misha} />
+    <Person {...olya} />
+    <Person {...alex} />
 
-    <section className="Person">
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
-  </div>
+  </>
 );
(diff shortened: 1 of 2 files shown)


mateusz-michalowski (changes requested):
obraz_2023-04-27_181333565
Well it doesn't pass the test, i think you should pass object into Person component not copy of an object also in return should be only section not all div with 'App' class. These two are only my thougths, but at the end of the day, it's not correct when automatic tests are not passed

src/App.jsx
```diff
@@ -24,22 +25,10 @@ export const alex = {
 };
 
 export const App = () => (
-  <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
+  <>
```
d-wasilewski, line 28: Instead of this fragment you should keep the div with className="App"

src/App.jsx
```diff
@@ -24,22 +25,10 @@ export const alex = {
...
+    <Person {...misha} />
+    <Person {...olya} />
```
d-wasilewski, line 30: it's better to just pass the entire object

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+const Person = ({ name, age, sex, isMarried, partnerName }) => {
+  const marriedSex = sex === 'f' ? 'husband' : 'wife';
```
d-wasilewski, line 4: checklist 2

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+
+  return (
+    <div className="App">
```
d-wasilewski, line 7: this should be in App.js, component should only have component-related code

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+            {`I am ${age}`}
+          </p>
+        )}
+
+        {!isMarried ? (
+          <p className="Person__partner">I am not married</p>
+        ) : (
+          <p className="Person__partner">
+            {isMarried ? `${partnerName} is my ${marriedSex}` : ''}
+          </p>
```
d-wasilewski, line 24: you return the same `p` tag twice, you can do this ternary check inside of this tag to not repeat yourself

d-wasilewski (changes requested):

d-wasilewski (commented):
also, fix the tests

d-wasilewski (changes requested):
nothing changed

src/App.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import Person from './components/Person/Person';
```
d-wasilewski, line 3: fix the tests, check how this component is imported there and adjust your code to that
image

d-wasilewski (approved):

PR #1243, opened 2023-04-26
Solution

KielpinskiJ:
[DEMO LINK](https://KielpinskiJ.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,22 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const married = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age ? (
+        <p className="Person__age">{`I am ${age}`}</p>
+      ) : (null)}
+      {isMarried ? (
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age ? (
+        <p className="Person__age">{`I am ${age}`}</p>
+      ) : (null)}
+      {isMarried ? (
```
bjeziorek, line 15: Nested ternary operator is a bad practice. Try to write this logic in another way.

bjeziorek (changes requested):

src/components/Person/Person.jsx
bjeziorek: Also when you use ternary operator line should start from "?" and ":" like this:
condition
  ? if_true
  : if_false

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age
+        ? <p className="Person__age">{`I am ${age}`}</p>
+        : null
+      }
```
kasiakubisiak, line 15: You can use `&&`

kasiakubisiak (changes requested):

kasiakubisiak (approved):

PR #1247, opened 2023-04-27
solution

KseniyaGonchar:
https://KseniyaGonchar.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,52 +1,36 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">
-      My name is
-      {' '}
-      {person.name}
-    </h2>
+export const Person = ({ person }) => {
+  const partnerSex = person.sex === 'm' ? 'wife' : 'husband';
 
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,57 @@
...
+
+    {person.isMarried === true && person.sex === 'm'
+        && (
+          <p className="Person__partner">
+            {person.partnerName}
+            {' '}
+            is my wife
+          </p>
+        )
+    }
```
qoa11a, line 30: Don't create so specific cases.

qoa11a (changes requested):
Good job! 👍 
But it could be done much easier. You could just create the variable for `partnerSex`.
It'd look like this: `const partnerSex = sex === 'm' ? 'wife' : 'husband'`.
Also, you could just check everything inside only one `p` tag.
It'd look like this: `<p className="Person__partner">{isMarried ? \`${partnerName} is my ${partnerSex}\` : 'I am not married'}</p>`.
Please, consider making the logic of the Person component easier.

src/components/Person/Person.jsx
KseniyaGonchar: thanks a lot!

PR #1248, opened 2023-04-27
Solution

mateusz-michalowski:
[DEMO LINK](https://mateusz-michalowski.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const marriedPrefix = sex === 'm' ? ' is my wife' : ' is my husband';
+  const marriedSuffix = isMarried ? partnerName + marriedPrefix : `I am not married`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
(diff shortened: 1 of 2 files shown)


SuaBy514 (approved):

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const marriedPrefix = sex === 'm' ? ' is my wife' : ' is my husband';
```
d-wasilewski, line 5: it's actually `suffix` to the name, you can figure out better names here than prefix and suuffix

d-wasilewski, line 5: ` is my x` is repeated here, you can just return the spouse here and combine it with the text in the `p` tag

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  const marriedSuffix = isMarried ? partnerName + marriedPrefix : `I am not married`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      <p className="Person__partner">{marriedSuffix}</p>
```
d-wasilewski, line 12: there should be a different text if a person is not married

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1276, opened 2023-05-15
solution

vova-sherstov:
[DEMO LINK](https://vova-sherstov.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.scss';
 
+import { Person } from './components/Person/Person';
+
 export const misha = {
   name: 'Misha',
   age: 37,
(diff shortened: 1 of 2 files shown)


src/App.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.scss';
 
+import { Person } from './components/Person/Person';
```
TarasHoliuk, line 4: no need to create a separate folder (Person) for a single file. A separate folder is required if the component (parent component) uses other (child) components related only to this parent component. But in this case `index.js` file should be added in the folder to re-export parent component

TarasHoliuk (changes requested):
Your demo link doesn't work. Make sure you followed the [instructions](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline). If you need help - feel free to ask in the chat

src/App.jsx
vova-sherstov: Demo link fixed.
I don't quite understand the comment about a separate folder for the person file. The person folder was already in the pool request. Do I need to put the person file in the component folder? like this "./components/Person"?

loralevitska (approved):
Well done)

PR #1288, opened 2023-05-25
https://grabynsky.github.io/react_person/

grabynsky:
https://grabynsky.github.io/react_person/

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,4 +11,4 @@ and use it 3 times inside the `App` instead of static markup.
 ## Instructions
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+    : `${person.partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+      {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+
+      <p className="Person__partner">
+        {person.isMarried ? `${personPartner}` : `I am not married`}
```
etojeDenys, line 15: No need to use template literals inside curly braces. Directly use the personPartner variable.

etojeDenys (changes requested):

src/components/Person/Person.jsx
grabynsky: Its conditional rendering. They must be enclosed in curly braсes. Does not work without curly brackets. I tried.  The personPartner variable checks whether you are male or female. And the condition checks marriage or not

OleziO (approved):

PR #1317, opened 2023-06-05
solution

SofiaLiubchenko:
[DEMO LINK](https://SofiaLiubchenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    <p className="Person__age">
+      {person.age ? `I am ${person.age}` : true}
+    </p>
+
+    <p>
+      {person.partnerName ? `${person.partnerName} is my ${person.sex === 'f' ? 'husband' : 'wife'}` : 'I am not married'}
+    </p>
+  </section>
+);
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    <p className="Person__age">
+      {person.age ? `I am ${person.age}` : true}
```
dinaminahi, line 10: Use logical operators && or || to conditionally return element

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+    </p>
+
+    <p>
+      {person.partnerName ? `${person.partnerName} is my ${person.sex === 'f' ? 'husband' : 'wife'}` : 'I am not married'}
```
dinaminahi, line 14: First of all checklist p.2
And create variable, so it will be easier to read this

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    <p className="Person__age">
+      {person.age ? `I am ${person.age}` : true}
+    </p>
+
+    <p>
```
dinaminahi, line 13: It should have className="Person__partner"

dinaminahi (changes requested):
Tests are failing, let's fix this

And678 (changes requested):
Checklist №2: `omit age if it is not given` is not implemented, that's why the tests are failing

PR #1351, opened 2023-06-10
add solution

VolodymyrPst:
https://volodymyrpst.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,18 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+function Person({ person }) {
```
VitaliyBondarenko1982, line 3: better to make destructuring of person for avoid to write each time

VitaliyBondarenko1982 (approved):
Well done

PR #1370, opened 2023-06-19
task solved

Nazarht:
[DEMO LINK](https://nazarht.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,20 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+    <p className="Person__age">
+      {person.age && `I am ${person.age}`}
+    </p>
+    <p className="Person__partner">
```
volodymyr-soltys97, line 11: If the person is not married, print the message 'I am not married'. Use a conditional expression to handle this case.

anastasiiavorobiova (approved):
Awesome!

PR #1387, opened 2023-06-27
Solution

yehor-kor:
[DEMO LINK](https://yehor-kor.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,39 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,39 @@
...
+          </p>
+        )
+      }
+
+      <p className="Person__partner">
+        {
+          isMarried
+            ? `${partnerName} is my `
+              + `${sex === 'm'
+                ? 'wife'
```
witflash, line 31: It's a bad practice you use nested ternary operators.
Try to fix it (you could separate logic with the both cases).

witflash (changes requested):
Good job!
Let's improve your code a little bit.

sTorba24 (approved):
GJ 🚀

PR #1392, opened 2023-06-27
Add task solution

vladmakohon:
[DEMO LINK](https://vladmakohon.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,30 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
+        )
+      }
+
+      <p className="Person__partner">
+        {
+          isMarried
+            ? (`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`)
```
ManoylovAC, line 24: using ternary operator inside another ternary operator is not good

for better readability you can create separate variable in component like `partnerStatus` and use it in jsx
```suggestion
  const partnerStatus = sex === 'm'
     ? 'wife'
     : 'husband';
     ...
     `${partnerName} is my ${partnerStatus}`
```

ManoylovAC (changes requested):
Very good start
Please check my comment about ternary operator and fix it

ManoylovAC (approved):
Good work

PR #1397, opened 2023-06-27
add task solution

NazarBaraban:
https://nazarbaraban.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,33 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+
+      {age >= 0 && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried ? (
+          `${partnerName} is my ${sex === 'm'
+            ? 'wife'
+            : 'husband'}`
```
ManoylovAC, line 24: you can create separate variable in component like `partnerStatus` and use it in jsx
```suggestion
  const partnerStatus = sex === 'm'
     ? 'wife'
     : 'husband';
     ...
     `${partnerName} is my ${partnerStatus}`
```

ManoylovAC (approved):
Good work
Please check my suggestion about additional variable

PR #1407, opened 2023-06-28
Solution

olena-ole:
[DEMO LINK](https://olena-ole.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -3,6 +3,9 @@ import './Person.scss';
 
 export const Person = ({ person }) => {
   const { name, age, sex, isMarried, partnerName } = person;
+  const spouse = sex === 'm'
+    ? 'wife'
+    : 'husband';
 
   return (
     <section className="Person">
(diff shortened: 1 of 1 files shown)


klmyana (approved):
GJ)

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age >= 0
+      && <p className="Person__age">{`I am ${age}`}</p>}
+
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${sex === 'm' ? 'wife' : 'husband'}`
```
ManoylovAC, line 18: Using ternary operator inside another one is not good idea

for better readability you can create separate variable in component like `partnerStatus` and use it in jsx
```suggestion
  const partnerStatus = sex === 'm'
     ? 'wife'
     : 'husband';
     ...
     `${partnerName} is my ${partnerStatus}`
```

ManoylovAC (changes requested):
Good work
Please check my suggestion about additional variable

tetiana-ivanchenko (approved):

PR #1426, opened 2023-06-28
added task solution

vlopatko:
[DEMO LINK](https://vlopatko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,33 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+
+      {isMarried && (
+        <p className="Person__partner">
+          {`${partnerName} is my ${hasMarriage}`}
+        </p>
+      )}
+
+      {!isMarried && (
```
dipinchuk, line 26: Good job!
When there are 2 variants for condition it is better to use `if-else`
You can get rid of duplicating this row 
```

dipinchuk (changes requested):

dipinchuk (approved):
Great!

tetiana-ivanchenko (approved):

PR #1428, opened 2023-06-28
solution

olegpaliy:
https://olegpaliy.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,19 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = (props) => {
+  const { name, age, sex, isMarried, partnerName } = props.person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+
+      <p className="Person__partner">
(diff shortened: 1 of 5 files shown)


TamaraShukalo (approved):
Good!✨

However, it is better to move the expression:
``` js
sex === 'f' ? 'husband' : 'wife'
```
to a separate constant to make the code more readable. 

```js
const partner = sex === 'f' ? 'husband' : 'wife';

File: src/components/Person/Person.jsx [hunk 1]
```diff
@@ -1 +1,19 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried && (`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`)}
+        {!isMarried && ('I am not married')}
```
And678, line 15: A perfect use-case for a ternary operator here:
```suggestion
        {
          isMarried
            ? (`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`)
            : ('I am not married')
        }
```

Not an impactful change for performance, but this way, we are evaluating `isMarried` only once, instead of twice

And678 (approved):
Approve :heavy_check_mark: , but check my comment below

PR #1429, opened 2023-06-28
Solution

outofs:
https://outofs.github.io/react_person/


src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = person.sex === 'f'
+    ? 'husband'
+    : 'wife';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
```
melnyksergey7, line 8: It`s better to declare variables like this: 
`const { name, age, sex, isMarried, partnerName } = person;`

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+
+      {person.age
+        && (<p className="Person__age">{`I am ${person.age}`}</p>)}
+
+      <p className="Person__partner">
+        {!person.isMarried
+          ? 'Im not married'
+          : `${person.partnerName} is my ${partner}`}
```
melnyksergey7, line 16: use only partnerName instead person.partnerName after declaring variables

melnyksergey7 (changes requested):

ManoylovAC (approved):
good work

PR #1445, opened 2023-06-30
develop

EvhenIvchenko:
[DEMO LINK](https://EvhenIvchenko.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
 // export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {` ${person.name}`}
```
VitaliyBodnar, line 6: it's not good practice to set the spaces this way, instead of you need to use margins and padding
check other places

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+        {` ${person.age}`}
+      </p>
+    )
+    }
+
+    {person.isMarried && (
+      <p className="Person__partner">
+        {`${person.partnerName} `}
+        is my
+        {` ${person.sex === 'm' ? 'wife' : 'husband'}`}
```
VitaliyBodnar, line 21: please create the constants for the 'm',  'wife', 'husband'
it will make a code cleaner

src/App.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
```
VitaliyBodnar, line 2: according to the style guide, import styles should be at the bottom of the imports list

VitaliyBodnar (changes requested):
Well done 🔥
Let's improve your solution 🧑‍💻

VitaliyBodnar (approved):

PR #1450, opened 2023-07-02
add person component

Misha408:
[DEMO LINK](https://misha408.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,25 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
```
SerhiiKirik, line 2: you can make a destructuring from person
```suggestion
export const Person = ({ person }) => {
  const { ... } = person;
  
  return (
    <section className="Person">
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried ? (
+        `${person.partnerName} is my ${person.sex === 'm' ? (
+          `wife`
+        ) : (
+          `husband`
```
SerhiiKirik, line 18: you can move this check to the variable

SerhiiKirik (changes requested):
Great 🔥

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1451, opened 2023-07-03
solution

vlad-fil:
[DEMO LINK](https://vlad-fil.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm'
+          ? 'wife'
+          : 'husband'}`
+        : 'I am not married'
+      }
```
IvanFesenko, line 19: Do not use nesting ternary operator
You may create variable that check wife/husband and use it

IvanFesenko (changes requested):
Good job. One improvement required

VitaliyBondarenko1982 (approved):
Well done

PR #1452, opened 2023-07-04
add task solution

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+
+    {person.age && (
+    <p className="Person__age">
+      {`I am ${person.age}`}
+    </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm'
```
anastasiiavorobiova, line 15: Avoid using magic strings it is considered programming antipattern, use constants with meaningful names instead

anastasiiavorobiova (changes requested):
Awesome work! Added a tiny comment to fix

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1483, opened 2023-07-16
react person

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+        }
+      </p>
+    )}
+    <p className="Person__partner">
+      {isMarried(person)}
+    </p>
+  </section>
+);
+
+function isMarried(person) {
```
etojeDenys, line 19:
```suggestion
function getMarriedLabel(person) {
```
When using the 'is' prefix in the function name 'isMarried,' we expect it to return either true or false

etojeDenys (changes requested):

OleziO (approved):

PR #1484, opened 2023-07-17
Develop

denis-kononenko:
[DEMO LINK](https://denis-kononenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried ? (
+        `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
```
anastasiiavorobiova, line 15: The 'sex' property is being directly compared to the string 'm'. It would be better to create constants for 'm' and 'f' to avoid magic strings and improve readability.

anastasiiavorobiova (changes requested):
Great work! Added a tiny comment to fix

anastasiiavorobiova (approved):
Awesome!

PR #1504, opened 2023-07-26
add task solution

vlad-dan1589:
https://vlad-dan1589.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <>
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      {person.age && (
+        <p className="Person__age">{`I am ${person.age}`}</p>
+      )}
+      <p className="Person__partner">
+        {person.isMarried ? (
+          `${person.partnerName} is my ${person.sex === 'f' ? ('husband') : ('wife')}`
+        ) : (
+          `I am not married`
+        )}
+      </p>
+    </section>
+  </>
+);
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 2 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
lerastarynets, line 1: Consider destructuring 'person' props in the parameter for a cleaner look.

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+  <>
```
lerastarynets, line 2: Remove the empty fragment tags, they are not necessary when returning a single element or component.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/components/Person/Person.jsx
Anastasiia-Svintsova: Not fixed 
image

Use destructuring for the person object.
```
const { name, age ... } = person;

File: src/components/Person/Person.jsx [hunk 3]
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && (
+      <p className="Person__age">{`I am ${person.age}`}</p>
+    )}
+    <p className="Person__partner">
+      {person.isMarried ? (
+        `${person.partnerName} is my ${person.sex === 'f' ? ('husband') : ('wife')}`
```
Anastasiia-Svintsova, line 9: Create a constant for a partner outside the JSX and move this logic there. The JSX code should be clean and focused on rendering components rather than defining complex logic directly within it.

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && (
+      <p className="Person__age">{`I am ${person.age}`}</p>
+    )}
+    <p className="Person__partner">
```
Anastasiia-Svintsova, line 7: Would be great to add spaces between multiline siblings for a better readability 
```suggestion
    )}
    
    <p className="Person__partner">
```

Anastasiia-Svintsova (changes requested):

src/components/Person/Person.jsx
vlad-dan1589: +

src/components/Person/Person.jsx
vlad-dan1589: +

src/components/Person/Person.jsx
vlad-dan1589: +

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  const partner = isMarried ? (
+    `${partnerName} is my ${sex === 'f' ? ('husband') : ('wife')}`
+  ) : (
+    `I am not married`
+  );
```
witflash, line 8: It's bad practice to use nested ternary operators.
Could you try to separate this logic? For example:
```
const partnerText = `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'`;
const statusText = isMarried ? partnerText : 'I am not married';
```

witflash (changes requested):
Good job!
Let's try to improve a little bit.

PR #1511, opened 2023-07-28
Solution

reseneweb:
https://reseneweb.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,25 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+    <p className="Person__partner">
+      {(() => {
+        if (person.isMarried) {
+          return person.sex === 'm'
+            ? `${person.partnerName} is my wife`
+            : `${person.partnerName} is my husband`;
+        }
+
+        return 'I am not married';
+      })()}
```
loralevitska, line 22: Avoid ternary nesting (even if you used `if` instead of one). Move this expression (person.sex === 'm'
            ? `${person.partnerName} is my wife`
            : `${person.partnerName} is my husband`;)  to variable, and then use interpolation
```suggestion
      {person.isMarried ? yourAwesomeVariableName : 'I am not married'}
```

loralevitska (changes requested):
A little more and it will be completed

lerastarynets (approved):

PR #1514, opened 2023-07-29
person done

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,31 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    {person.isMarried === true && person.sex === 'm' && (
+      <p className="Person__partner">
+        {`${person.partnerName} is my wife`}
+      </p>
+    )}
+
+    {person.isMarried === true && person.sex === 'f' && (
+      <p className="Person__partner">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    {person.isMarried === true && person.sex === 'm' && (
```
lerastarynets, line 13: Avoid using strict equality and instead use truthy check when checking for 'true' or 'false'. This is because sometimes values might be truthy or falsy but not strictly equal to true or false.
```suggestion
    {person.isMarried && person.sex === 'm' && (
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+      <p className="Person__partner">
+        {`${person.partnerName} is my wife`}
+      </p>
+    )}
+
+    {person.isMarried === true && person.sex === 'f' && (
```
lerastarynets, line 19: same
```suggestion
    {person.isMarried && person.sex === 'f' && (
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+      <p className="Person__partner">
+        {`${person.partnerName} is my husband`}
+      </p>
+    )}
+
+    {person.isMarried === false && (
```
lerastarynets, line 25:
```suggestion
    {!person.isMarried && (
```

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

lerastarynets (approved):

PR #1521, opened 2023-08-01
practical task 3

Andrii-Monko:


src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+    <p className="Person__partner">
+      {`${person.partnerName} is my husband`}
+    </p>
+    )}
+
+    { person.hasOwnProperty('partnerName') && person.sex === 'm' && (
+    <p className="Person__partner">
+      {`${person.partnerName} is my wife`}
+    </p>
+    )}
```
loralevitska, line 32: Marriage text move to variable and combine tags into 1. It should be looks like this 
```suggestion
    <p className="Person__partner">
        {person.isMarried
          ? `${person.partnerName} is my ${yourAwesomeVariableName}`
          : 'I am not married'}
      </p>
    )}
```

PR #1523, opened 2023-08-02
Develop

Hanna-Balabukha:
https://Hanna-Balabukha.github.io/react_person/


src/App.jsx
```diff
@@ -1,6 +1,21 @@
 import React from 'react';
+import { Person } from './Person';
 import './App.scss';
 
+export const App = () => (
+  <div className="App">
+    <Person
+      person={misha}
+    />
```
Anastasiia-Svintsova, line 9: There is no need to split it into several lines when there are fewer than 3 props
```suggestion
    <Person person={misha} />
```

src/Person.jsx
```diff
@@ -0,0 +1,36 @@
...
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried,
+    partnerName = 'I am not married' } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {' '}
+        {name}
```
Anastasiia-Svintsova, line 12: Consider using [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for better readability

src/Person.jsx
```diff
@@ -0,0 +1,36 @@
...
+      </h2>
+      <p className="Person__age">
+        {age && (
```
Anastasiia-Svintsova, line 15: Don't render this element if there is no age 
```suggestion
      {age && (
        <p className="Person__age">
```

src/Person.jsx
```diff
@@ -0,0 +1,36 @@
+import './App.scss';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried,
+    partnerName = 'I am not married' } = person;
```
Anastasiia-Svintsova, line 5: The partnerName should be a name. You can create a separate variable for a partner and assign a value here 
```suggestion
  const { 
    name, 
    age, 
    sex, 
    isMarried,
    partnerName
  } = person;
```

Anastasiia-Svintsova (changes requested):
image
image

lerastarynets (approved):

PR #1525, opened 2023-08-02
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,20 +1,28 @@
 /* eslint-disable react/jsx-one-expression-per-line */
 export const Person = ({ person }) => {
-  const spouse = person.sex === 'f'
+  const {
+    name,
+    age,
+    partnerName,
+    isMarried,
+    sex,
+  } = person;
+
+  const spouse = sex === 'f'
     ? 'husband'
     : 'wife';
 
   return (
     <section className="Person">
-      <h2 className="Person__name">My name is {person.name}</h2>
+      <h2 className="Person__name">My name is {name}</h2>
       {person.age
-        ? <p className="Person__age">I am {person.age}</p>
+        ? <p className="Person__age">I am {age}</p>
         : null
         }
-      {person.isMarried
+      {isMarried
         ? (
           <p className="Person__partner">
-            {`${person.partnerName} is my ${spouse}`}
+            {`${partnerName} is my ${spouse}`}
           </p>
         )
         : <p className="Person__partner">I am not married</p>


src/components/Person/Person.jsx
```diff
@@ -1 +1,40 @@
-// export const Person = ({ person }) => ();
+/* eslint-disable react/jsx-one-expression-per-line */
+export const Person = ({ person }) => {
```
darokrk, line 2: think you can use destructing props to get values shorter and more readable e.g.:

```
  const {
    name,
    age,
    partnerName,
    isMarried,
    sex,
  } = person;
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,40 @@
...
+  if (person.sex === 'f') {
```
darokrk, line 3: think you don't need this if else check, components looks the same so you can only check what sex person has and they pass it to paragraph with partenr info: e.g:

Add this statement before return ()
```
 const spouse = sex === 'f'
    ? 'husband'
    : 'wife';
```

And your partner info can look like this:

```
  {isMarried ? (
          {`${partnerName} is my ${spouse}`}
        </p>
      ) : (
        <p className="Person__partner">I am not married</p>
      )}
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,40 @@
...
+    return (
+      <section className="Person">
+        <h2 className="Person__name">My name is {person.name}</h2>
+        {typeof (person.age) === 'number'
```
darokrk, line 7: don't need to check if age is type number, think it is redundant here so you can remove this ternary operator

src/components/Person/Person.jsx
```diff
@@ -1 +1,40 @@
...
+              {person.partnerName} is my husband
+            </p>
+          )
+          : <p className="Person__partner">I am not married</p>
+        }
+      </section>
+    );
+  }
+
+  return (
```
darokrk, line 23: that code can be removed

darokrk (changes requested):
Few mentions in the code need to be fixed 🛠️

src/components/Person/Person.jsx
Luk2asz: If I use destructuring  props I must to change my code in APP.jsx and then tests do not pass

DorotaLeniecDev (approved):
Good job!

DorotaLeniecDev (approved):
Approved

PR #1538, opened 2023-08-05
solution

taras-nakonechnyi:
https://taras-nakonechnyi.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,27 @@
-// export const Person = ({ person }) => ();
+import cn from 'classnames';
+
+export const Person = ({ person: {
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+} }) => (
+  <section className={cn('Person')}>
+    <h2 className={cn('Person__name')}>{`My name is ${name}`}</h2>
+    {age && (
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+import cn from 'classnames';
+
+export const Person = ({ person: {
+  name,
+  age,
+  sex,
+  isMarried,
+  partnerName,
+} }) => (
+  <section className={cn('Person')}>
```
VitaliyBodnar, line 10: library classnames need to use for making dynamic classnames, in this case, you have a static classname
so, `cn` is redundant here

src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+    <p className={cn('Person__partner')}>
+      {isMarried ? (
+        `${partnerName} is my ${sex === 'm' ? (
+          'wife'
+        ) : (
+          'husband'
+        )}`
+      ) : (
+        'I am not married'
+      )}
```
VitaliyBodnar, line 24: it's a bad practice to use nested ternary, better to move this code to the variable

VitaliyBodnar (changes requested):
Almost done!

PR #1552, opened 2023-08-11
all tests pass

aniashev:
[DEMO LINK](https://aniashev.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,27 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${person.name}`}
+      </h2>
+
+      { person.age !== undefined && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+      )}
+
+      <p className="Person__partner">
+        {!person.isMarried
+          ? `I am not married`
+          : `${person.partnerName} is my ${
+            person.sex === 'f'
+              ? `husband`
+              : `wife`
+          }`}
+      </p>
+    </section>
+  </div>
+);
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <div className="App">
```
anastasiia-tilikina, line 4: you don't need to wrap in `App` here
```suggestion
  
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+      )}
+
+      <p className="Person__partner">
+        {!person.isMarried
+          ? `I am not married`
+          : `${person.partnerName} is my ${
+            person.sex === 'f'
+              ? `husband`
+              : `wife`
+          }`}
```
anastasiia-tilikina, line 23: using nested ternary operators is bad practice
Let's move this logic to a separate method

anastasiia-tilikina (changes requested):
GJ!
let's just fix small issues

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+    { person.age !== undefined && (
+    <p className="Person__age">
+      {`I am ${person.age}`}
+    </p>
+    )}
+
+    <p className="Person__partner">
+      {!person.isMarried
+        ? `I am not married`
+        : `${person.partnerName} is my ${getTitle(person)}`}
```
VitaliyBodnar, line 18: please don't call a function from the template because it will be executed every time when the state change, so make sense to move it to the variable

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+    </p>
+  </section>
+);
+
+const getTitle = person => (person.sex === 'f' ? `husband` : `wife`);
```
VitaliyBodnar, line 23: if you want to use this method in a component, you need to declare this method inside this component

VitaliyBodnar (changes requested):
Almost done!

VitaliyBodnar (approved):
Good job!

PR #1559, opened 2023-08-14
react person

alina-isakova:
[DEMO LINK](https://alina-isakova.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <>
+      <h2 className="Person__name">
+        {`My name is ${person.name}`}
+      </h2>
+
+      {person.age && (
+        <p className="Person__age">
+          {`I am ${person.age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+          {`I am ${person.age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {person.isMarried === true ? (
+          `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+        ) : (
+          'I am not married'
+        )}
```
Viktor-Kost, line 19: 1) move this conditions to a variable to increase readability
2) `person.isMarried === true ?` is the same as just `person.isMarried ?`

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
```
Viktor-Kost, line 2: it's a good practice to use object destructing

```suggestion
export const Person = ({ person }) => {
  const {...} = person
  
  return (
    <section className="Person">
```

Viktor-Kost (changes requested):

lerastarynets (approved):

PR #1562, opened 2023-08-15
Solution

92BS:
[DEMO LINK](https://92BS.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,22 +1,22 @@
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">
-      My name is
-      {' '}
-      {person.name}
-    </h2>
-    {person.age && (
-      <p className="Person__age">
-        I am
-        {' '}
-        {person.age}
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerGender = isMarried ? (`${partnerName} is my ${sex === 'm' ? ('wife') : ('husband')}`) : (`I am not married`);
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {partnerGender}
       </p>
-    )}
-    <p className="Person__partner">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {' '}
+      {person.name}
```
Anastasiia-Svintsova, line 6: Consider using [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for better readability

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+    </h2>
+    {person.age && (
+      <p className="Person__age">
+        I am
+        {' '}
+        {person.age}
```
Anastasiia-Svintsova, line 12: Same

src/components/Person/Person.jsx
Anastasiia-Svintsova, line 6: Use destructuring for objects for better code readability

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+      </p>
+    )}
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'f'
+          ? 'husband' : 'wife'}`
```
Anastasiia-Svintsova, line 18: Create a variable for a partner and move calculations with a ternary operator there

Anastasiia-Svintsova (changes requested):
Good start! Check my comments, please

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerGender = isMarried ? (`${partnerName} is my ${sex === 'm' ? ('wife') : ('husband')}`) : (`I am not married`);
```
Anastasiia-Svintsova, line 3: Would be better to leave only this logic in this variable and use the conditional rendering in JSX 
```suggestion
  const partnerGender = sex === 'm' ? 'wife' : 'husband';
```

Anastasiia-Svintsova (approved):

PR #1569, opened 2023-08-17
Solution

oksana-vas:
[DEMO LINK](https://oksana-vas.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,17 +1,29 @@
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">My name is {person.name}</h2>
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const isMarried = () => {
```
mykhalenych, line 2:
```suggestion
  const showPartnerInfo = () => {
```

mykhalenych (approved):

src/components/Person/Person.jsx
oksana-vas: thank you for the comment !

PR #1586, opened 2023-08-21
add task solution

valet2102:


Moroz-Dmytro (changes requested):
GJ 👍 but you need to fix tests

P.S.
image

PR #1589, opened 2023-08-21
add task solution

Wesses:
[DEMO LINK](https://Wesses.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,47 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+const SEX_FEMALE = 'f';
+const SEX_MALE = 'm';
+
+function getPartnerInformation(sex, partnerName) {
+  if (sex === SEX_FEMALE) {
+    return (
+      <p className="Person__partner">{`${partnerName} is my husband`}</p>
+    );
+  }
+
+  if (sex === SEX_MALE) {
+    return (
+      <p className="Person__partner">{`${partnerName} is my wife`}</p>
+    );
+  }
+
+  return null;
+}
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    sex,
+    age,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {typeof age === 'number' && (
+        <p className="Person__age">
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+    return (
+      <p className="Person__partner">{`${partnerName} is my husband`}</p>
+    );
+  }
+
+  if (sex === SEX_MALE) {
+    return (
+      <p className="Person__partner">{`${partnerName} is my wife`}</p>
+    );
+  }
```
Moroz-Dmytro, line 17: Maybe we can simplify this. The only difference between these 2 returns is the partner `wife` or `husband`. So better to use ternary here

src/components/Person/Person.jsx
```diff
@@ -1 +1,47 @@
...
+    age,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {typeof age === 'number' && (
```
Moroz-Dmytro, line 35:
```suggestion
      {!!age && (
```

Moroz-Dmytro (changes requested):
Some comments below

src/components/Person/Person.jsx
Wesses: if age === 0 ?

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+
+      {!!age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      { isMarried
+        ? (<p className="Person__partner">{`${partnerName} is my ${getPartnerStatus(sex)}`}</p>)
+        : (<p className="Person__partner">I am not married</p>)
```
polosanya, line 28: You can put this condition inside "Person__partner" element, as it's used in both cases

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+    age,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {!!age && (
```
polosanya, line 20: It's not critical, but would be great to check if the age is a positive number

polosanya (approved):
Good job!

PR #1592, opened 2023-08-21
Solution

Oleh0293:
[DEMO LINK](https://Oleh0293.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,25 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name,
+    age,
+    sex,
+    isMarried,
+    partnerName } = person;
+  const partnerDesignation = sex === 'm'
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {person.age && (<p className="Person__age">{`I am ${age}`}</p>)}
+      <p className="Person__partner">
+        { isMarried
+          ? `${partnerName} is my ${partnerDesignation}`
+          : 'I am not married'
+        }
+      </p>
+    </section>
+  );
+};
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name,
+    age,
+    sex,
+    isMarried,
+    partnerName } = person;
```
polosanya, line 8:
```suggestion
  const { 
    name,
    age,
    sex,
    isMarried,
    partnerName,
  } = person;
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+  const partnerDesignation = sex === 'm'
```
polosanya, line 9:
```suggestion
  const partner = sex === 'm'
```
You can simplify the variable's name

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {person.age && (<p className="Person__age">{`I am ${age}`}</p>)}
```
polosanya, line 16: It's not critical, but you could check if the age is a positive number

polosanya, line 16:
```suggestion
      <h2 className="Person__name">
        {`My name is ${name}`}
      </h2>
      
      {person.age && (
        <p className="Person__age">
          {`I am ${age}`}
        </p>
      )}
```
Such formatting will make your code easier to read

polosanya (changes requested):
Almost perfect!

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {person.age && (
```
tiserett, line 21: why use person.age here? just use age ;)

btw, I guess if age is gonna be 0 it's not gonna work correctly, I mean age 0 is not correctly itself, but react will probably just render '0' on the page, so we have to convert it to boolean to make sure it works just fine, we can achieve it by using !! or Boolean

PR #1594, opened 2023-08-21
solution

Capoverde:
[DEMO LINK](https://Capoverde.github.io/react_person/)

Something wrong with a server, that's why it is not with a tests.


choeqq (approved):
Looks good to me ✅

PR #1595, opened 2023-08-21
add task solution

kaavka:
[DEMO](https://kaavka.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,31 @@
-// export const Person = ({ person }) => ();
+const SEX_MALE = 'm';
+
+export const Person = ({
+  person: { name, age, sex, isMarried, partnerName },
+}) => {
+  const partnerInfo = sex === SEX_MALE ? (
+    `${partnerName} is my wife`
+  ) : (
+    `${partnerName} is my husband`
+  );
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {typeof age === 'number' && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? (partnerInfo)
+          : `I am not married`}
+      </p>
+    </section>
+  );
+};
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
-// export const Person = ({ person }) => ();
+const SEX_MALE = 'm';
+
+export const Person = ({
+  person: { name, age, sex, isMarried, partnerName },
```
Moroz-Dmytro, line 4: It's difficult to read. Better to do this destructuring inside function body

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+      {typeof age === 'number' && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? (partnerInfo)
+          : `I am not married`}
```
Moroz-Dmytro, line 27:
```suggestion
        {isMarried
          ? (partnerInfo)
          : `I am not married`
        }
```

Moroz-Dmytro (changes requested):
GJ 👍 
Some comments below

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+    return 'I am not married';
+  };
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {typeof age === 'number' && (
```
Anna-Tkachenko, line 20: Will be better to use `Boolean()` function instead of checking the type of the argument
We'll check the variable type further using Typescript
Let's check the value here

Also for now if age will be equal to 0, it will show the result on the page

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {getPartnerInfo()}
```
Anna-Tkachenko, line 27: It's bad practice to use functions inside `jsx`
As it's hard to predict what function will return and what result user will see on the page
Let's avoid function here
Just prepare the content based on variables before the return statement

Anna-Tkachenko (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? (partnerInfo)
+          : `I am not married`
+        }
```
dffUqp, line 23:
```suggestion
        {isMarried
          ? partnerInfo
          : `I am not married`
        }
```

dffUqp (approved):
GJ✔️

PR #1602, opened 2023-08-22
Solution

desevoker:
[DEMO LINK](https://desevoker.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const partnerType = (person.sex === 'm') ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+      {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+
+      <p className="Person__partner">
+        {person.isMarried ? (
+          `${person.partnerName} is my ${partnerType}`
+        ) : (
+          'I am not married'
+        )}
+      </p>
+    </section>
+  );
+};
--- a/README.md
+++ b/README.md
@@ -12,4 +12,4 @@ and use it 3 times inside the `App` instead of static markup.
 ## Instructions
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_person/) and add it to the PR description.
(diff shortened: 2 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const partnerType = (person.sex === 'm') ? 'wife' : 'husband';
```
BudnikOleksii, line 4: Let's move the magic value to a constant

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+      {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
```
BudnikOleksii, line 10: You can use destructuring to get values so you don't need to write every time `person.age` and etc

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+
+      <p className="Person__partner">
+        {person.isMarried ? (
+          `${person.partnerName} is my ${partnerType}`
+        ) : (
+          'I am not married'
+        )}
```
BudnikOleksii, line 17:
```suggestion
        {person.isMarried
          ? `${person.partnerName} is my ${partnerType}`
          : 'I am not married'
        }
```

BudnikOleksii (changes requested):
Good job!
Left a few comments, check them

src/components/Person/Person.jsx
desevoker: But the syntax I used is taken from the React docs and was also used by Misha.

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partnerType = (sex === SEX_MALE) ? 'wife' : 'husband';
```
Anna-Tkachenko, line 14:
```suggestion
  const partnerType = sex === SEX_MALE
    ? 'wife'
    : 'husband';
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+
+  const partnerText = isMarried
+    ? `${partnerName} is my ${partnerType}`
+    : 'I am not married';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {!!age && <p className="Person__age">{`I am ${age}`}</p>}
```
Anna-Tkachenko, line 24: Move paragraph to the new line, let's make code clearer
```suggestion
      {!!age && (
        <p className="Person__age">
          {`I am ${age}`}
        </p>
      )}
```

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1604, opened 2023-08-22
react_person

maximmorenko:
[DEMO LINK](https://maximmorenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,25 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {person.name}
+    </h2>
+    {person.age && (
+      <p className="Person__age">
+        I am
+        {person.age}
+      </p>
+    )}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName}
+        is my
+        {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
+  </section>
+);
--- a/README.md
+++ b/README.md
@@ -12,4 +12,4 @@ and use it 3 times inside the `App` instead of static markup.
 ## Instructions
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_person/) and add it to the PR description.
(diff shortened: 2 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
```
Viktor-Kost, line 4: it's a good practice to use object destructing
```suggestion
export const Person = ({ person }) => {
  const {...} = person;
  
  return (
    <section className="Person">
      ...
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+    )}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName}
+        is my
+        {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
```
Viktor-Kost, line 23: here is you should change only the text inside `<p>` tag.

1. move condition into a variable `const partnerStatus = sex === 'm' ? 'wife' : 'husband'`
2. use like this. Example:

```suggestion
      <p className="Person__partner">{isMarried ? `${partnerName} is my ${partnerStatus}` : 'I am not married'</p>
```

Viktor-Kost (changes requested):
Nice!
Some minor changes needed

src/components/Person/Person.jsx
maximmorenko: <p className="Person__partner">{isMarried ? `${partnerName} is my ${partnerStatus}` : 'I am not married'</p>
this is not working
Parsing error: Unterminated regular expression
may be so
<p className="Person__partner">{person.isMarried ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}` : 'I am not married'}</p>

Viktor-Kost: yes. since as i asked at previous comment you need to use object destructing first.
Then create a variable `const partnerStatus = sex === 'm' ? 'wife' : 'husband'`

Viktor-Kost (changes requested):
Added an answer to your comment. If you need any help feel free to ask me at fe_chat

src/components/Person/Person.jsx
maximmorenko: may be destructing is a good practice to use object, but the condition of this task is not require this. isn't it? if i do not use destructing is it a mistake?

Viktor-Kost: no, it's not a mistake. It's a good practice that developers apply at real projects. During your studying you'll have requirements yet mentors don't only check requirements. They're here to help to achieve something more. Especially with improving your solutions

Viktor-Kost (approved):
Nice!

PR #1607, opened 2023-08-22
Develop

edgergard:
[DEMO LINK](https://edgergard.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,31 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const SEX_MALE = 'm';
+  const isWifeOrHusband = sex === SEX_MALE
+    ? `${partnerName} is my wife`
+    : `${partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && <p className="Person__age">{age && `I am ${age}`}</p>}
+      <p className="Person__partner">
+        {isMarried
+          ? isWifeOrHusband
+          : `I am not married`
+        }
+      </p>
+    </section>
+  );
+};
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const SEX_MALE = 'm';
+  const isWifeOrHusband = sex === SEX_MALE
```
Anna-Tkachenko, line 13: I suppose that in this kind of variable should be stored some boolean value
If you need to store the marriage data, will be better to use the name according to this statement
`marriageInfo` will be better name

src/components/Person/Person.jsx
```diff
@@ -1 +1,31 @@
...
+    ? `${partnerName} is my wife`
+    : `${partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && <p className="Person__age">{age && `I am ${age}`}</p>}
```
Anna-Tkachenko, line 22: This code is not readable, let's move the paragraph to the new line
```suggestion
      {age && (
        <p className="Person__age">{age && `I am ${age}`}</p>
      )}
```

Anna-Tkachenko: Also it's not clear why did you check `age` variable 2 times here
Inner check is redundant, delete it

Anna-Tkachenko (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+    ? `${partnerName} is my wife`
+    : `${partnerName} is my husband`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
```
Moroz-Dmytro, line 23:
```suggestion
        <p className="Person__age">
          {`I am ${age}`}
        </p>
```

Moroz-Dmytro (approved):

PR #1608, opened 2023-08-22
add task solution

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
 // export const Person = ({ person }) => ();
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const sexPartner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried
+        ? (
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const sexPartner = sex === 'm' ? 'wife' : 'husband';
```
Anna-Tkachenko, line 3:
```suggestion
  const sexPartner = sex === 'm'
    ? 'wife'
    : 'husband';
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried
+        ? (
+          <p className="Person__partner">{`${partnerName} is my ${sexPartner}`}</p>
```
Anna-Tkachenko, line 11: Don't repeat yourself: the outer paragraph is the same for both cases
Change just the content depending on the variable value

Anna-Tkachenko (changes requested):

Moroz-Dmytro (approved):

PR #1610, opened 2023-08-22
add task solution

Vannnkof:
[DEMO LINK](https://vannnkof.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,33 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+const SEX_MALE = 'm';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+
+      {isMarried ? (
+        <p className="Person__partner">
+          {`${partnerName} is my ${sex === SEX_MALE
+            ? 'wife'
+            : 'husband'}`}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
+    </section>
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {age && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
```
tiserett, line 20: I guess if age is gonna be 0 it's not gonna work correctly, I mean age 0 is not correctly itself, but react will probably just render '0' on the page, so we have to convert it to boolean to make sure it works just fine, we can achieve it by using !! or Boolean

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+
+      {isMarried ? (
+        <p className="Person__partner">
+          {`${partnerName} is my ${sex === SEX_MALE
+            ? 'wife'
+            : 'husband'}`}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
tiserett, line 30: let's write ternary inside of <p> tag

+ let's create variable to get rid of inner ternary

tiserett (changes requested):

Anna-Tkachenko (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+
+      {Boolean(age) && (
+        <p className="Person__age">{`I am ${age}`}</p>
+      )}
+
+      <p className="Person__partner">
+        {isMarried
+          ? messageAboutPartner
+          : messageAboutFreedom}
+      </p>
```
roman-mirzoian, line 32:
```suggestion
      <p className="Person__partner">
        {isMarried
          ? messageAboutPartner
          : messageAboutFreedom
         }
      </p>
```

roman-mirzoian (approved):
Well done 👍

PR #1614, opened 2023-08-22
solution

OlhaMomot:


src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+    }
+
+    <p className="Person__partner">
+      {!person.isMarried && ('I am not married')}
+      {person.isMarried && (
+        person.sex === MALE_SIGN ? (
+          `${person.partnerName} is my wife`
+        ) : (
+          `${person.partnerName} is my husband`
+        ))}
```
dffUqp, line 24: it can be replace with ternary like 
```jsx
            {person.isMarried
              ? ...
              : ...}
```

If you don't want to use ternary inside ternary try to create separate variable

dffUqp (approved):
Almost done:smiley:

PR #1616, opened 2023-08-22
add task solution

cty2802:


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+    <>
+      {person.age && (
```
Anna-Tkachenko, line 8: Let's get needed properties from `person` object using destructuring before the return statement

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+    </>
+
+    <p className="Person__partner">
+      {person.isMarried && person.sex === 'm' && (
+        <>{`${person.partnerName} is my wife`}</>
+      )}
+      {person.isMarried && person.sex === 'f' && (
+        <>{`${person.partnerName} is my husband`}</>
+      )}
+      {!person.isMarried && <>I am not married</>}
```
Anna-Tkachenko, line 22: You've repeated some part of code
Will be better to define the info about the partner initially, should it be `wife` or `husband`
And then use this data while preparing the content

Anna-Tkachenko (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const WOMAN_GENDER = 'f';
+  const MAN_GENDER = 'm';
```
roman-mirzoian: Such constants are better defined outside the component.

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+    }
+
+    return <>I am not married</>;
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      <>
```
roman-mirzoian, line 28: Do we really need to wrap everything in a fragment here?

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+    name,
+    age,
+    isMarried,
+    sex,
+    partnerName,
+  } = person;
+
+  function getMarriedInfo() {
+    if (sex === MAN_GENDER && isMarried) {
+      return <>{`${partnerName} is my wife`}</>;
```
roman-mirzoian: Do we really need to wrap everything in a fragment here?

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+    }
+
+    if (sex === WOMAN_GENDER && isMarried) {
+      return <>{`${partnerName} is my husband`}</>;
```
roman-mirzoian: Do we really need to wrap everything in a fragment here?

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+    }
+
+    return <>I am not married</>;
```
roman-mirzoian, line 21: Do we really need to wrap everything in a fragment here?

src/components/Person/Person.jsx
```diff
@@ -1 +1,44 @@
...
+
+  const {
+    name,
+    age,
+    isMarried,
+    sex,
+    partnerName,
+  } = person;
+
+  function getMarriedInfo() {
```
roman-mirzoian, line 14: Such a function can also be taken outside the component definition.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/components/Person/Person.jsx
```diff
@@ -1 +1,39 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+const MAN_GENDER = 'm';
+
+function getMarriedInfo(status, gender, companion) {
```
roman-mirzoian, line 5: It is worth choosing more meaningful names for arguments so that they explain what is happening in the code.
```suggestion
function getMarriedInfo(isMarried, gender, partnerName) {
```

roman-mirzoian (approved):
Well done, but pay attention to the comment.

BudnikOleksii (approved):
Good job!

PR #1620, opened 2023-08-22
add task solution

KlymSudakov:


src/App.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
+import { Person } from './components/Person/Person';
 import './App.scss';
 
 export const misha = {
   name: 'Misha',
-  age: 37,
+  age: '37',
```
Anna-Tkachenko, line 7: Why did you change the type for the variable? It should be number, not string

src/components/Person/Person.jsx
```diff
@@ -1 +1,34 @@
...
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const MALE = 'm';
+  const notMarried = 'I am not married';
+  const personPartner = (sex === MALE)
```
Anna-Tkachenko, line 12: `partnerStatus` will be better variable name

src/components/Person/Person.jsx
```diff
@@ -1 +1,34 @@
...
+    ? 'wife'
+    : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
```
Anna-Tkachenko, line 18:
```suggestion
      <h2 className="Person__name">
        {`My name is ${name}`}
      </h2>
```

Anna-Tkachenko (approved):

PR #1621, opened 2023-08-22
add task solution

yuliiaant:
[DEMO LINK](https://yuliiaant.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,11 +1,11 @@
 import './Person.scss';
 
 export const Person = ({ person }) => {
-  const partnerSex = person.sex === 'f'
+  const partnerInfo = person.sex === 'f'
     ? 'husband'
     : 'wife';
-  const personsPartner = person.isMarried
-    ? `${person.partnerName} is my ${partnerSex}`
+  const marriageStatusInfo = person.isMarried
+    ? `${person.partnerName} is my ${partnerInfo}`
     : 'I am not married';
 
   return (
@@ -21,7 +21,7 @@ export const Person = ({ person }) => {
       )}
 
       <p className="Person__partner">
-        {personsPartner}
+        {marriageStatusInfo}
       </p>
     </section>
   );


src/App.jsx
```diff
@@ -25,21 +26,16 @@ export const alex = {
...
 export const App = () => (
   <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
+    <Person
+      person={misha}
+    />
```
Anna-Tkachenko, line 31: If component accept only one property, it's not required to move this property to the new line
```suggestion
    <Person person={misha} />
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const partnerSex = person.sex === 'f'
+    ? 'husband'
+    : 'wife';
+  const personIsMarried = person.isMarried
```
Anna-Tkachenko, line 7: For now you've added variable name the same as the properties name
So it's not clear what is the difference between `isMarried` property and `personIsMarried` variable

Anna-Tkachenko (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const partnerSex = person.sex === 'f'
```
Anna-Tkachenko, line 4: Will be better to use `partnerInfo`

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+    ? 'husband'
+    : 'wife';
+  const personsPartner = person.isMarried
```
Anna-Tkachenko, line 7: It's not persons partner
Will be better to use `marriageStatusInfo`

Anna-Tkachenko (approved):

PR #1627, opened 2023-08-22
Solution🎀

polinavafik:
[DEMO LINK](https://polinavafik.github.io/react_person/)

we could change this `  const partner = sex === 'm'
    ? 'wife'
    : 'husband';` 
 to  just using the word 'spouse'  here
```
        { isMarried
          ? `${partnerName} is my spouse`
          : 'I am not married'
        }
      </p>
```

so it would work for all couples 😎👍

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,39 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  const partner = sex === 'm'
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,39 @@
...
+      </h2>
+
+      {age > 0
+        ? (
+          <p className="Person__age">
+            {`I am ${age}`}
+          </p>
+        )
+        : ''
+      }
```
Anna-Tkachenko, line 29: If you don't need to show something in case if age < 0, use `&&` operator instead of ternary one

Anna-Tkachenko (changes requested):

dffUqp (approved):
GJ✔️

PR #1651, opened 2023-08-23
solution

zvir91:
[DEMO LINK](https://zvir91.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,35 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+  let whoIsPartner = '';
+
+  if (isMarried && sex === 'm') {
+    whoIsPartner = `${partnerName} is my wife`;
+  } else if (isMarried && sex === 'f') {
+    whoIsPartner = `${partnerName} is my husband`;
+  } else {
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+  } = person;
+  let whoIsPartner = '';
+
+  if (isMarried && sex === 'm') {
+    whoIsPartner = `${partnerName} is my wife`;
+  } else if (isMarried && sex === 'f') {
+    whoIsPartner = `${partnerName} is my husband`;
+  } else {
+    whoIsPartner = 'I am not married';
+  }
```
maxim2310, line 17: not so good to make this logic directly in component body, better to make it inside funk and call it when you need

maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed

src/components/Person/Person.jsx
```diff
@@ -1 +1,38 @@
...
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {
+        age
+         && (
+           <p className="Person__age">
+             {`I am ${age}`}
+           </p>
+         )
+      }
```
maxim2310, line 32:
```suggestion
        age && (
           <p className="Person__age">
             {`I am ${age}`}
           </p>
         )
      }
```

maxim2310 (approved):
Good work!!!

PR #1661, opened 2023-08-25
Blindfold Solution

neovendano:


neovendano:
[DEMO LINK](https://neovendano.github.io/react_person/)

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+    : NOT_MARRIED_TEXT;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {
+        age && (
```
roman-mirzoian, line 24: You have to explicitly state that `age` it's a boolean, because it can just render 0.

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+const FEMALE_SIGN = 'f';
+
+const NOT_MARRIED_TEXT = 'I am not married';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  const partnerStatus = sex === FEMALE_SIGN
+    ? 'husband'
+    : 'wife';
```
roman-mirzoian, line 12: These strings could also be transferred to separate constants.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

BudnikOleksii (approved):
Good work 👍

PR #1664, opened 2023-08-25
react_person

annak1104:


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${
+          person.sex !== 'm' ? 'husband' : 'wife'
+        }`}
```
Moroz-Dmytro, line 15: It's difficult to read. Create variable for this value `person.sex !== 'm' ? 'husband' : 'wife'`

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+    {person.isMarried ? (
+      <p className="Person__partner">
```
Moroz-Dmytro, line 12: Remember about DRY. This tag should be the same for both variants, so you need to change only content depending on `isMarried` status

Moroz-Dmytro (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+
+const Person = ({ person }) => {
+  const { name, age, partnerName, sex, isMarried } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && (
```
roman-mirzoian, line 11: It is worth converting this value to boolean manually, because if there is 0, it will simply render 0:
```suggestion
      {!!age && (
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${sex !== PERSON_MAN_SEX ? 'husband' : 'wife'}`
```
roman-mirzoian, line 18: `husband` and `wife` could also be cast as constants.

roman-mirzoian (approved):
Well done, but pay attention to the comments.

PR #1671, opened 2023-08-28
Solution

Denis-Sizemov:
https://Denis-Sizemov.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,25 +1,29 @@
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">
-      {`My name is ${person.name}`}
-    </h2>
-    {person.age && (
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+        if (!person.isMarried) {
+          return 'I am not married';
+        }
+
+        if (person.isMarried && person.sex === 'm') {
+          return `${person.partnerName} is my wife`;
+        }
+
+        return `${person.partnerName} is my husband`;
+      })()}
```
etojeDenys, line 22: in jsx we trying to not use IIFE, so let's move all these conditions in some vars in our component body

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1678, opened 2023-08-31
task solution was added

Liliya-Kalinichenko:
[DEMO LINK](https://Liliya-Kalinichenko.github.io/react_person/)


anastasiia-tilikina (changes requested):
GJ!

let's just make a small fix

PR #1694, opened 2023-09-11
Solution

KarolinaWiktoria:
[DEMO](https://KarolinaWiktoria.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerMale = (sex === 'm') ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {person.age && <p className="Person__age">I am {age}</p>}
```
darokrk, line 12: Instead of 'person.age', you can just use 'age' since it has already been destructured at the start of the component.

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+      {(isMarried)
```
darokrk, line 13: this logic can be simplified by, e.g.:

<p className="Person__partner">{isMarried ? `${partnerName} is my ${partnerMale}` : 'I am not married'}</p>

darokrk (changes requested):
Looks good 😊  left comments for improvements 🛠️

plodnikate (approved):
Good job :fire:!

PR #1702, opened 2023-09-18
Solution

oksom:
DEMO LINK: https://oksom.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, isMarried, partnerName, sex } = person;
+
+  return (
+    <section className="Person">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+      </h2>
+      {age !== undefined && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${sex === 'm' ? 'wife' : 'husband'}`
+          : 'I am not married'}
```
VitaliyBodnar, line 17: nested ternary operators is a bad practice, it's better to move this parts of code to the variables

VitaliyBodnar (changes requested):

maxim2310 (approved):
Good work!!!

PR #1709, opened 2023-09-22
Solution IrynaMP

IrynaMP:
IrynaMP


volodymyr-soltys97 (changes requested):
There is no the demo link, also you need to add the demo link to the PR description
image

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+  const getPartnerLabel = () => {
+    let marriedLabel;
+
+    if (isMarried) {
+      marriedLabel = sex === 'm'
+        ? `${partnerName} is my wife`
+        : `${partnerName} is my husband`;
+    }
+
+    return isMarried === true
```
TarasHoliuk, line 13:
```suggestion
    return isMarried
```
isMarried is a boolean value, so no need to compare it with `true`, moreover in the ternary the variables in the condition part will be automatically transformed into boolean (the same as in the `if` block condition) - FYI

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+      : 'I am not married';
+  };
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age !== undefined
```
TarasHoliuk, line 24:
```suggestion
      {!!age
```
you may just transform it into boolean - FYI

TarasHoliuk (approved):
The previous comments were fixed - approved ✔️ 
Added 2 comments FYI - no need to re-request review, the task is approved

PR #1710, opened 2023-09-23
Solution

GeorgeSavostikov:
[DEMO LINK](https://GeorgeSavostikov.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,22 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${person.name}`}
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+
+      {person.age && (
+        <p className="Person__age">
+          {`I am ${person.age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {person.isMarried
+          ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
```
VitaliyBodnar, line 16: Using nested ternary operators is a bad practice. It's better to move it to the variable

VitaliyBodnar (changes requested):

VitaliyBodnar (approved):

PR #1741, opened 2023-10-04
solution

OlhaKoziuk:
[DEMO LINK](https://OlhaKoziuk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      {person.age && (
+        <p className="Person__age">{`I am ${person.age}`}</p>
+      )}
+      {person.isMarried ? (
+        <p className="Person__partner">{`${person.partnerName} is my husband`}</p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
+    </section>
+  </div>
+);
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${name}`}</h2>
+    {age && (
+      <p className="Person__age">{`I am ${age}`}</p>
+    )}
+    {isMarried ? (
+      <p className="Person__partner">{`${partnerName} is my husband`}</p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
```
sviatikkondur, line 16: I think you can create a single `p` element and set the value depending on the value of isMarried

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person: {
+  name,
+  age,
+  isMarried,
+  partnerName,
+} }) => (
```
sviatikkondur, line 6: To pass tests try to destructure these field like in checklist:
```jsx
export const Component = (props) => {
 const {
  firstProperty,
  secondProperty,
  thirdProperty
} = props; 

 return (
  <>
    <h1>{firstProperty}</h1>
    <span>{secondPropery}</span>
    <p>{thirdProperty}</p>
  </>
 )
};
```

sviatikkondur (changes requested):
This is optional, but you can also create a styles file for the Person component and set all styles there

ArtemTeslenko (approved):
Greate!

PR #1745, opened 2023-10-04
solution react person task

ivaZaiets:
[DEMO LINK](https://ivaZaiets.github.io/react_person/)


vlevik (changes requested):
Good job, but check linter, please
image

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+} }) => (
+  <>
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age === undefined ? (
+        null
+      ) : (
```
ArtemTeslenko, line 16: it is bad practice to compare with undefined, also no need to use ternary here

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+      <p className="Person__partner">
+        {!isMarried ? (
+          'I am not married'
+        ) : (
+          `${partnerName} is my ${sex === 'f' ? (
+            'husband'
+          ) : (
+            'wife'
+          )}`
+        )}
```
ArtemTeslenko, line 31: nested ternary is not a good idea, create variable and move there this inside condition

ArtemTeslenko (changes requested):
Nice! Read one more time about conditional rendering, be careful))

src/components/Person/Person.jsx
```diff
@@ -1 +1,37 @@
...
+  )}`;
+
+  return (
+    <>
+      <section className="Person">
+        <h2 className="Person__name">
+          {`My name is ${name}`}
+        </h2>
+
+        {typeof age !== 'undefined' && (
```
ArtemTeslenko, line 21: just `{age && (<p>....}`

ArtemTeslenko (changes requested):
Good! One more comment. Repeat about `&&`, `??`, `||` operators, it is necessary to understand how they work))

ArtemTeslenko (approved):
Very good🔥

PR #1748, opened 2023-10-04
add react person solution

Kuper18:
Hi, here is my solution
[DEMO LINK](https://kuper18.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,39 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  function getPartnerSex(personSex) {
+    return personSex === 'f' ? 'husband' : 'wife';
+  }
+
+  const marriedStatus = isMarried
+    ? `${partnerName} is my ${getPartnerSex(sex)}`
+    : 'I am not married';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,39 @@
...
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  function getPartnerSex(personSex) {
```
dsfreedom, line 10: great solution to not overuse conditional operators in the following code!)

src/components/Person/Person.jsx
```diff
@@ -1 +1,39 @@
...
+
+  const marriedStatus = isMarried
+    ? `${partnerName} is my ${getPartnerSex(sex)}`
+    : 'I am not married';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
```
dsfreedom, line 22: Here bettween tags declarations you could've put some spaces so it is a bit easier to read. Hope it make sense)
<opening tag
  something 
 <closing tag

{ age something
 }
 
 {isMarried
 }

src/components/Person/Person.jsx
```diff
@@ -1 +1,39 @@
...
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+      {isMarried ? (
```
dsfreedom, line 28: I believe conditional operator here is redundant as you do it before in marriedStatus, so here you can use p tag only once

dsfreedom (approved):
Great job! Only few small remarks

src/components/Person/Person.jsx
Kuper18: Good point)

ArtemTeslenko (approved):
Good job! There are small issues, fix please)

PR #1749, opened 2023-10-04
add task solution

dsfreedom:
[DEMO LINK](https://dsfreedom.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,30 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({
+  person: {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  },
+}) => (
+  <section className="Person">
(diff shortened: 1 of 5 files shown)


YevhenKharko (approved):
Perfect! Nice solution

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+
+    {age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {isMarried
+        ? `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`
```
ArtemTeslenko, line 25: nested ternary - is not a good idea, move inside condition into some variable, here just use this variable

ArtemTeslenko (changes requested):
Good job! Do not use nested ternary)

polosanya (approved):
Well done!

PR #1755, opened 2023-10-04
add task solution

annabasenyuk:
[DEMO LINK](https://annabasenyuk.github.io/react_person/)
Hi) Added the task solution, check it please)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
@@ -25,21 +26,8 @@ export const alex = {
 
 export const App = () => (
   <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person person={misha} />
+    <Person person={olya} />
+    <Person person={alex} />
   </div>
 );
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const maritalStatus = () => {
+    let status = '';
+
+    if (!person.isMarried) {
+      status = 'I am not married';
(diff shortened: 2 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+    if (!person.isMarried) {
+      status = 'I am not married';
+    } else if (person.sex === 'f') {
+      status = `${person.partnerName} is my husband`;
+    } else {
+      status = `${person.partnerName} is my wife`;
+    }
+
+    return status;
+  };
```
BetterSol, line 14: You should consider changing this part of the code according to the checklist
```
1. [CODE STYLE] - Use destructuring for getting access to values of `props` object```

File: src/components/Person/Person.jsx [hunk 2]
```diff
@@ -1 +1,29 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      { person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
```
BetterSol, line 21: same here, `CODE STYLE 1`

BetterSol, line 21: better to use a ternary operator like in `CODE STYLE 2` and `CODE STYLE 3` (optional)

BetterSol (changes requested):
Good Job! Put a few comments to make your code look even better:)

src/App.jsx
```diff
@@ -25,21 +26,8 @@ export const alex = {
...
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person person={misha} />
```
vasylmiskiv, line 29: It's not necessary but you could use `map` method instead of repeating the same component for example create array of the objects

src/components/Person/Person.jsx
```diff
@@ -1 +1,37 @@
...
+  const { name,
+    age,
+    sex,
+    isMarried,
+    partnerName } = person;
+
+  const maritalStatus = () => {
+    let status = '';
+
+    if (!isMarried) {
```
vasylmiskiv, line 11: There's a lot of unnecessary code here because it doesn't read well and pretty tough to maintain

You can check that if the person is unmarried, it returns "not married"
Otherwise you check with a ternary operator for the gender of the person and return the correct partner status with the name of the partner

vasylmiskiv (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,33 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      { age && (
+      <p className="Person__age">
+        {`I am ${age}`}
+      </p>
```
StasSokolov, line 24: check formatting

StasSokolov (approved):
Great job🔥

PR #1760, opened 2023-10-05
add React person task solution

o-drozzdyk:
[DEMO LINK](https://o-drozzdyk.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
@@ -25,21 +26,8 @@ export const alex = {
 
 export const App = () => (
   <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person person={misha} />
+    <Person person={olya} />
+    <Person person={alex} />
   </div>
 );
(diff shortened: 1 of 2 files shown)


tetyana-pol (approved):

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
vasylmiskiv, line 1: You need to descructure properties from the object

src/App.jsx
```diff
@@ -25,21 +26,8 @@ export const alex = {
...
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person person={misha} />
```
vasylmiskiv, line 29: Also you could use `map` method instead of repeating the same component

vasylmiskiv (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+          {`I am ${age}`}
+        </p>
+      )}
+
+      <p className="Person__partner">
+        {!isMarried
+          ? `I am not married`
+          : `${partnerName} is my ${sex === 'm'
+            ? `wife`
+            : `husband`}`}
```
ArtemTeslenko, line 28: nested ternary is not a good idea, to determine wife or husband you can move it into variable and here just use this variable

src/App.jsx
```diff
@@ -23,23 +24,10 @@ export const alex = {
...
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    {people.map(person => <Person person={person} />)}
```
ArtemTeslenko, line 31: `map` is from the next topic, but if you decided to use it, you should know about `key`, here you can use `persone.name` as `key`

ArtemTeslenko (changes requested):
Good job! Try to avoid of using nested ternary))

vasylmiskiv (approved):

PR #1762, opened 2023-10-05
solution

MVoutch:


src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
```
YaroslavYarynych, line 2: You should destructurIng our person by name, age, sex etc. Like this: const { a, b } = obj;
It`s really hard to read code without it.
You can read about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

YaroslavYarynych (changes requested):
Good job, but you should correct one point and It will be fine

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+        {isMarried
+          ? (
+            <p>
+              { `${partnerName} is my ` }
+              { sex === 'm' ? 'wife' : 'husband' }
+            </p>
+          ) : (
+            <p>I am not married</p>
+          )
+        }
```
superpooperxxx, line 32: This is really hard to read
1. Let's calculate this logic outside of JSX in some variable `sex === 'm' ? 'wife' : 'husband'`
2. You don't need these inner `<p>`, just
```
  {isMarried
    ? `${partnerName} is my ${calculatedLogic}`
    :  'I am not married'
  }
</p>

superpooperxxx (changes requested):
Good job🔥 Let's make the solution more clean

superpooperxxx (approved):
🔥🔥🔥

PR #1764, opened 2023-10-05
Solution

YaroslavYarynych:
[DEMO LINK](https://YaroslavYarynych.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,5 +1,13 @@
+import './Person.scss';
+
 export const Person = ({ person }) => {
-  const { name, age, sex, partnerName } = person;
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
 
   return (
     <section className="Person">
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+      }
+
+      <p className="Person__partner">
+        {!isMarried
+          ? 'I am not married'
+          : `${partnerName} is my ${sex === 'm'
+            ? 'wife'
+            : 'husband'}`
+        }
+      </p>
```
ArtemTeslenko, line 32: Try to avoid nested ternary, it is better to use variable to determine `wife` or `husband`, also create constant for message if person is not married. For example `const NOT_MARRIED_MESSAGE = '...'`

ArtemTeslenko (changes requested):
Nice! Just one moment, nested ternary is a bad practice, try to avoid it)

superpooperxxx (approved):
Perfection🔥🤌🏻

PR #1769, opened 2023-10-06
add task solution

AndriiMysheniuk:
[DEMO LINK](https://andriimysheniuk.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
@@ -11,6 +12,7 @@ export const misha = {
 
 export const olya = {
   name: 'Olya',
+  age: undefined,
   sex: 'f',
   isMarried: true,
   partnerName: 'Maksym',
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,45 @@
...
+        {' '}
+        {name}
+      </div>
+      {age && (
+      <div className="Person__age">
+        I am
+        {' '}
+        {age}
+      </div>
+      )}
```
etojeDenys, line 41:
```suggestion
      {age && (
      <div className="Person__age">
        {`I am ${age}`}
      </div>
      )}
```
we can simplify it

src/components/Person/Person.jsx
```diff
@@ -1 +1,45 @@
...
+        <div className="Person__partner">
+          {partnerName}
+          {' '}
+          is my husband
+        </div>
+      );
+    }
+  } else {
+    partner = <div className="Person__partner">I am not married</div>;
+  }
```
etojeDenys, line 26: let's just save wife/husband in some var and use it in return instead of this logic

etojeDenys (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  let partnerStatus = null;
+
+  if (isMarried) {
+    partnerStatus = sex === 'm' ? 'wife' : 'husband';
+  } else {
+    partnerStatus = 'I am not married';
+  }
```
VitaliyBodnar, line 10: make sense to use ternary operator here as it will make a code shorter

VitaliyBodnar (changes requested):

src/App.jsx
```diff
@@ -11,6 +12,7 @@ export const misha = {
 
 export const olya = {
   name: 'Olya',
+  age: undefined,
```
etojeDenys, line 15: we don't need to specify this property

src/App.jsx
```diff
@@ -21,25 +23,13 @@ export const alex = {
   age: 25,
   sex: 'm',
   isMarried: false,
+  partnerName: undefined,
```
etojeDenys, line 26: same here

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #1782, opened 2023-10-14
Solution Person

MyroslavSynenko:
DEMO LINK https://myroslavsynenko.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,22 +1,25 @@
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">
-      {`My name is ${person.name}`}
-    </h2>
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
 
-    {person.age && (
-      <p className="Person__age">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried
+        && (person.sex === 'm'
+          ? `${person.partnerName} is my wife`
+          : `${person.partnerName} is my husband`)}
+
+      {!person.isMarried && 'I am not married'}
```
VitaliyBondarenko1982, line 19: suggest to make more readable and shorter this part.
before return jsx in component we can decide which text will be rendered.
also suggest to destructure person;
```
const personPartner = sex === 'm' ? 'wife' : 'husband';
const personPartnerText = isMarried ? `${person.partnerName} is my ${personPartner}` : 'I am not married';
```
```suggestion
      {personPartnerText}
```

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comment.

anastasiiavorobiova (approved):
Great work!

PR #1810, opened 2023-10-29
Solution

momos1703:
[DEMO LINK](https://momos1703.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,20 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    sex,
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried
+        ? <p className="Person__partner">{`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`}</p>
+        : <p className="Person__partner">I am not married</p>
+      }
+    </section>
+  );
+};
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+    isMarried,
+    partnerName,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried
+        ? <p className="Person__partner">{`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`}</p>
```
volodymyr-soltys97, line 17: Consider will be better if you create a variable for this check and use it here
{`${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`}

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/components/Person/Person.jsx
momos1703: Good idea! Done.

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+    : 'wife';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried
+        ? <p className="Person__partner">{`${partnerName} is my ${partnerRole}`}</p>
+        : <p className="Person__partner">I am not married</p>
+      }
```
volodymyr-soltys97, line 23: Consider it will be better and more readable, because we do not duplicate the tag with the same class twice, but write the condition inside immediately
```suggestion
        <p className="Person__partner">
          {isMarried ? `${partnerName} is my ${partnerRole}` : 'I am not married'}
        </p>
```

volodymyr-soltys97 (changes requested):
Almost done!

src/components/Person/Person.jsx
momos1703: I agree. Done!

volodymyr-soltys97 (approved):
Well done 🔥

PR #1820, opened 2023-11-03
Develop

TenetaJ:


src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+    {person.age > 0 && (
+      <p className="Person__age">
+        I am
+        {` ${person.age}`}
```
etojeDenys, line 9:
```suggestion
        {`I am ${person.age}`}
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,36 @@
...
+        ))
+          || (person.sex === 'f' && (
+            <p className="Person__partner">
+              {
+                `${person.partnerName} is my husband`
+              }
+            </p>
+          ))
+      )}
+  </section>
```
etojeDenys, line 35: let's move some part of this logic to var, it should improve the readability

etojeDenys (changes requested):

src/components/Person/Person.jsx
TenetaJ: done

src/components/Person/Person.jsx
TenetaJ: done

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let partnerMessage = '';
+
+  if (person.isMarried === false) {
```
VitaliyBodnar, line 4: It's better to use switch case statement instead of multiple if else blocks

VitaliyBodnar (changes requested):

src/components/Person/Person.jsx
TenetaJ: done

PR #1828, opened 2023-11-06
mySolution

AntonFof:
[DEMO LINK](https://AntonFof.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  let partnerText = 'I am not married';
+
+  if (isMarried) {
+    if (sex === 'm') {
+      partnerText = `${partnerName} is my wife`;
+    } else {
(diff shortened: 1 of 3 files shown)


nazarmatsevych (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
```
anastasiia-tilikina, line 11: age always will be in your object. you don't need to check

anastasiia-tilikina (changes requested):
just one tiny issue

src/components/Person/Person.jsx
AntonFof: Olya doesn't have the property 'age'. What is necessary to do?

VitaliyBodnar: okay, for this case you can use this condition

VitaliyBodnar (approved):

PR #1845, opened 2023-11-15
react_person

KolomietsN:
[DEMO LINK](https://KolomietsN.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,24 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+        {`I am ${age}`}
+      </p>
+      )}
+      {isMarried ? (
+        <p className="Person__partner">
+          {`${partnerName} is my ${partner}`}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
volodymyr-soltys97, line 21: Now this tag with the same class is duplicated 2 times, try to fix it, and check already inside the tag

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #1846, opened 2023-11-15
add person component

yefimov77:
[DEMO LINK](https://yefimov77.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = (
+  { person: { name, age, sex, isMarried, partnerName } },
+) => {
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age ? (<p className="Person__age">{`I am ${age}`}</p>) : ''}
+      <p className="Person__partner">{isMarried ? (`${partnerName} is my ${partner}`) : 'I am not married'}</p>
+    </section>
+  );
+};
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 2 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = (
+  { person: { name, age, sex, isMarried, partnerName } },
+) => {
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age ? (<p className="Person__age">{`I am ${age}`}</p>) : ''}
```
Anton-Kuchmasov, line 9: bad practice, Misha said about it in one of pre-tasks videos :)

rather use:
`   {age ? (
<p className="Person__age">
{`I am ${age}`}
</p>) : ''}`

Anton-Kuchmasov, line 9: better way is using && 
f.e. age && <p className...>

if age === false, tag <p> even shouldn't be render and add to DOM 


i had some problems with tests for it, so be careful

Anton-Kuchmasov (changes requested):
nice job! 

but give it some format, not in one string with html,js and working with props :)

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
-// export const Person = ({ person }) => ();
+export const Person = (
+  { person: { name, age, sex, isMarried, partnerName } },
```
Anton-Kuchmasov, line 2: nice! 
for my solution, i forgot about destruct:)

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+) => {
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && (
```
Anton-Kuchmasov, line 9: i'm approving this, but please,
avoid nested ternary (i have some simular issue with my code right now)

Anton-Kuchmasov (approved):

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
...
+      <p className="Person__age">{`I am ${age}`}</p>
+      )}
+      <p className="Person__partner">
+        {isMarried
```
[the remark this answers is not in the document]
yefimov77: > this is not critical, but it is better to move logic from the layout to separate functions like this:
> 
> ```
> const getPartner = () => {
>   if (isMarried) {
>     return `${partnerName} is my ${partner}`;
>   }
>   return 'I am not married';
> };
> 
> <p className="Person__partner">{getPartner()}</p>
> ```

so, i`d better create such function, i see. But, anyway i leave them in the same component file, but outside function, that i export?

PR #1851, opened 2023-11-16
Solution

BoblykNazar:
[DEMO LINK](https://BoblykNazar.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.scss';
 
+import { Person } from './components/Person/Person';
+
 export const misha = {
   name: 'Misha',
   age: 37,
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,40 @@
...
+        {`My name is ${name}`}
+      </h2>
+
+      {age && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      {(isMarried && sex === 'm') && (
```
mykhailoyurkovskyi, line 23: I guess it is better to create a function like this and do not hardcode
    const getPartner = () => {
      if (isMarried) {
        return `${partnerName} is my ${partner}`;
      }

      return 'I am not married';
  };

mykhailoyurkovskyi (changes requested):
Great job, but in my mind code needs some improvements👌

src/App.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.scss';
 
+import { Person } from './components/Person/Person';
+
 export const misha = {
```
ArtemTeslenko, line 6: you can move people into separate file, it will make your component cleaner

src/components/Person/Person.jsx
```diff
@@ -1 +1,34 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const {
+    name,
+    age,
+    isMarried,
+    sex,
+    partnerName,
+  } = person;
+  const partner = sex === 'f' ? 'husband' : 'wife';
```
ArtemTeslenko, line 9: try to avoid magic values `const FIMALE = 'f'`
```suggestion
  const partner = sex === FIMALE ? 'husband' : 'wife';
```

ArtemTeslenko (approved):
Great 🔥 See comments)

PR #1878, opened 2023-11-21
Task is done

YuriySobakar:
[DEMO LINK](https://YuriySobakar.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```
BoblykNazar, line 2: You can format as here:
const { 
  name,
  age, 
  sex, 
  isMarried,
  partnerName } = person;

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+  const FEMALE = 'f';
+
+  return (
+    <>
+      <section className="Person">
+        <h2 className="Person__name">{`My name is ${name}`}</h2>
+        {age
```
BoblykNazar, line 9: Try using here '&&' instead of the ternary operators. Then there haven't to return 'null'

BoblykNazar (approved):
well done!💪

src/App.jsx
```diff
@@ -25,21 +26,16 @@ export const alex = {
...
 export const App = () => (
   <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
+    <Person
+      person={misha}
+    />
```
BudnikOleksii, line 31:
```suggestion
    <Person  person={misha} />
```
You have only one prop and it fits in one line, and other too

src/components/Person/Person.jsx
BudnikOleksii: > Try using here '&&' instead of the ternary operators. Then there haven't to return 'null'

Agree, it doesn't make sense to use ternary here

BudnikOleksii (changes requested):
Good job 👍 
Just a few minor fixes required

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const FEMALE = 'f';
```
BudnikOleksii, line 3: Constants should be placed outside the component

BudnikOleksii (approved):
Good job 👍

PR #1879, opened 2023-11-21
solution for react_person

Anton201220:
https://Anton201220.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,25 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`}
+      </p>
+    ) : (
+      <p className="Person__partner">
+        I am not married
+      </p>
+    )}
```
etojeDenys, line 23: let's move some part of this logic to var, it will simplify the readability

etojeDenys (changes requested):

nazarmatsevych (approved):
Great work!

PR #1880, opened 2023-11-22
complete task

Annbpiu:
[DEMO LINK](https://Annbpiu.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,19 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
```
nazarmatsevych, line 15: Move this part to separate variable partner:
`const partner = person.sex === 'm' ? 'wife' : 'husband'; and use it here to avoid nested ternary

nazarmatsevych (changes requested):
Good work, let's make a small improvement

nazarmatsevych (approved):
Awesome 👍

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
-// export const Person = ({ person }) => ();
+export const Person = ({
+  person,
+  partner = person.sex === 'm' ? 'wife' : 'husband',
```
nazarmatsevych, line 3: But it's not a prop, next time move component variables below and add return statement to render html

Annbpiu: you mean 
export const Person = ({
  person,
  partner = person.sex === 'm' ? 'wife' : 'husband';
  
  return (...)
  });?

PR #1881, opened 2023-11-22
react_person_solution

vadimkondratiuk:
[DEMO LINK](https://vadimkondratiuk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,32 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,32 @@
...
+        (isMarried && (
+          <p className="Person__partner">
+            {
+              sex === 'm'
+                ? `${partnerName} is my wife`
+                : `${partnerName} is my husband`
+            }
+          </p>
+        ))
+        || <p className="Person__partner">I am not married</p>
```
VitaliyBodnar, line 28: It's better to use ternary operator instead of && with ||
As it will make your code cleaner

VitaliyBodnar (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #1882, opened 2023-11-22
Solution

ArtemVelychko:
[DEMO LINK](https://ArtemVelychko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,28 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      My name is
+      {' '}
+      {person.name}
+    </h2>
+
+    {person.age
+      ? (
+        <p className="Person__age">
+          I am
+          {' '}
+          {person.age}
+        </p>
+      ) : null}
+
+    {person.isMarried
+      ? (
+        <p className="Person__partner">
+          {person.sex === 'm'
+            ? `${person.partnerName} is my wife`
+            : `${person.partnerName} is my husband`}
+        </p>
+      )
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+    </h2>
+
+    {person.age
+      ? (
+        <p className="Person__age">
+          I am
+          {' '}
+          {person.age}
+        </p>
+      ) : null}
```
anastasiia-tilikina, line 16: use interpolation.
when you need to render smth only if condition is true it is better to use And (&&)  logical operator instead of ternary with the second option as null.

```javascript
{perosn.age && (
// your jsx goes here
)
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+
+    {person.isMarried
+      ? (
+        <p className="Person__partner">
+          {person.sex === 'm'
+            ? `${person.partnerName} is my wife`
+            : `${person.partnerName} is my husband`}
+        </p>
+      )
+      : (<p className="Person__partner">I am not married</p>)}
```
anastasiia-tilikina, line 26: do not use nested ternary operators. move logic into separate var or helper.

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
anastasiia-tilikina, line 1: it is also a good idea to use destructuring for person

anastasiia-tilikina (changes requested):
GJ!

let's make some improvements

anastasiia-tilikina (approved):
Good job!

Just check my comment. we don't need conditionally render the whole element, jus its content

PR #1891, opened 2023-11-27
solution

jacekkosiba:
[DEMO LINK](https://<your_account>.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,11 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,11 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${person.sex === 'f' ? 'husband' : 'wife'}`}
+      </p>
+    ) : <p className="Person__partner">I am not married</p>}
```
zaura333, line 9: You could consider deconstructing person for clarity, e.g.:
` const { name, age, isMarried, sex, partnerName } = person;`

zaura333 (approved):
Looks good, just one suggestion

src/components/Person/Person.jsx
mvjl000: I would say that in this case it's just a matter of preference. I use destructuring on objects only when I want to access some value more than once... or when it simply looks better ;)

mvjl000 (approved):
Well done!

PR #1898, opened 2023-11-28
solution

MarcinLigmann:
https://MarcinLigmann.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export function Person({ person }) {
+  let typeOfPartner = 'wife';
+
+  if (person.sex === 'f') {
+    typeOfPartner = 'husband';
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age ? <p className="Person__age">I am {person.age}</p> : null}
+      {person.isMarried
+        ? <p className="Person__partner">{person.partnerName} is my {typeOfPartner}</p>
+        : <p className="Person__partner">I am not married</p>}
+    </section>
+  );
+}
--- a/README.md
+++ b/README.md
@@ -12,4 +12,4 @@ and use it 3 times inside the `App` instead of static markup.
 ## Instructions
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://<your_account>.github.io/react_person/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://MarcinLigmann.github.io/react_person/) and add it to the PR description.
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@mate-academy/cypress-tools": "^1.0.4",
     "@mate-academy/eslint-config-react": "^0.0.11",
     "@mate-academy/eslint-config-react-typescript": "*",
-    "@mate-academy/scripts": "^1.2.8",
+    "@mate-academy/scripts": "^1.2.12",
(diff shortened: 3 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1,16 +1,32 @@
...
+      <h2 className="Person__name">
+        My name is
+        {name}
+      </h2>
+      {age ? (
+        <p className="Person__age">
+          I am
+          {age}
+        </p>
+      ) : null}
```
Patrykmclaren14, line 21: `null` here is unnecessary

src/components/Person/Person.jsx
```diff
@@ -1,16 +1,32 @@
-export function Person({ person }) {
+export function Person({ person: { name, age, sex, isMarried, partnerName } }) {
   let typeOfPartner = 'wife';
 
-  if (person.sex === 'f') {
+  if (sex === 'f') {
     typeOfPartner = 'husband';
   }
```
Patrykmclaren14: You can change unneeded code lines to one line with a ternary operator
```
const spouse = sex === 'm' ? 'wife' : 'husband';
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        My name is
+        {' '}
+        {name}
+      </h2>
+      {age ? (
+        <p className="Person__age">
```
mbruhler, line 16: Use short-circuting `&&`

src/components/Person/Person.jsx
mbruhler: yes, true

src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+          I am
+          {' '}
+          {age}
+        </p>
+      ) : null}
+      {isMarried
+        ? (
+          <p className="Person__partner">
```
mbruhler, line 24: Why u do `{' '}`? Use backtick string {`${partnerName} is my ${typeOfPartner}`}

src/components/Person/Person.jsx
mbruhler: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

mbruhler (changes requested):
Your code needs some changes

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
-// export const Person = ({ person }) => ();
+export function Person({ person: { name, age, sex, isMarried, partnerName } }) {
+  const spouse = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+      {age && (
```
mbruhler, line 9: In fact I said it wrong. You should check for `!!age` as if `age === 0`, it will output `0` to the template.

mbruhler (changes requested):
I'm sorry but I made a mistake.

Please refer to the comments below.

NOTE: Resolve comments after changes by clicking "Resolve comment" button.

src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
...
+      )}
+
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {`${partnerName} is my ${spouse}`}
+          </p>
+        )
+        : <p className="Person__partner">I am not married</p>
+      }
```
mvjl000, line 22: I would do the ternary operator inside of `p` tag and only change text inside of it.

What benefits may come from this? For example: When the className for the tag changes, you don't have to correct it in two places 

Something like this:
```suggestion
<p className="Person__partner">
      {isMarried
        ? `${partnerName} is my ${spouse}`
        : 'I am not married'
      }
      </p>
```

mvjl000 (changes requested):

mbruhler (approved):
Well done, approved

PR #1917, opened 2023-12-01
solution

aureliagorzna:
- [DEMO LINK](https://aureliagorzna.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,31 @@
-// export const Person = ({ person }) => ();
+/* eslint-disable react/jsx-one-expression-per-line */
+export function Person({ person }) {
+  let marriage = 'I am not married';
+
+  if (person.isMarried) {
+    if (person.sex === 'm') {
+      marriage = `${person.partnerName} is my wife`;
+    }
+
+    if (person.sex === 'f') {
+      marriage = `${person.partnerName} is my husband`;
+    }
+  }
+
+  if ('age' in person) {
+    return (
+      <section className="Person">
+        <h2 className="Person__name">My name is {person.name}</h2>
+        <p className="Person__age">I am {person.age}</p>
+        <p className="Person__partner">{marriage}</p>
+      </section>
+    );
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      <p className="Person__partner">{marriage}</p>
+    </section>
+  );
+}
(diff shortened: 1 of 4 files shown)


Timi2424 (changes requested):
you need to deploy your page

mvjl000 (changes requested):
I highly recommend you to look for some open-source react projects, to see how people write logic in there

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+  }
+
+  if (person.isMarried && person.sex === 'f') {
+    marriage = `${person.partnerName} is my husband`;
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {'age' in person
```
JoyCoffeeAddict, line 16: if age is undefined, the condition is still true, but we don't want to display this.

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+  let marriage = 'I am not married';
+
+  if (person.isMarried && person.sex === 'm') {
+    marriage = `${person.partnerName} is my wife`;
+  }
+
+  if (person.isMarried && person.sex === 'f') {
+    marriage = `${person.partnerName} is my husband`;
+  }
+
```
JoyCoffeeAddict, line 12: Simplify as you did before
you can write it with one `if` statement and `ternary` operator.

JoyCoffeeAddict (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+      : marriage = `${person.partnerName} is my husband`;
+  }
+
+  if (person.isMarried && person.sex === 'm') {
+    marriage = `${person.partnerName} is my wife`;
+  }
+
+  if (person.isMarried && person.sex === 'f') {
+    marriage = `${person.partnerName} is my husband`;
+  }
```
mbruhler, line 18: I don't understand this.
Why u do the same operation twice? Can you elaborate?

Looks like in 6-10 you assign `${person.partnerName} is my husband / wife`; 

and then you do the same

mbruhler (changes requested):
Please elaborate

src/components/Person/Person.jsx
aureliagorzna: I forgot to remove the old code

mbruhler (approved):

PR #1918, opened 2023-12-03
Solution

vasylynaBC:


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <div className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+    {person.age !== undefined && (
```
VitaliyBondarenko1982, line 6:
```suggestion
    {person.age && (
```
it can be shorter

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <div className="Person">
```
VitaliyBondarenko1982, line 2: suggest to destructure person

```suggestion
export const Person = ({ person }) => {
const { // destructure here } = person;
return (
  <div className="Person">
```

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comments.

lerastarynets (approved):

PR #1924, opened 2023-12-04
react_person

TVChulkova:
[DEMO LINK](https://TVChulkova.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,24 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age !== undefined
+      && (
+        <p className="Person__age">
+          {`I am ${person.age}`}
+        </p>
+      )
+    }
+
+    <p className="Person__partner">
+      {person.isMarried === true
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+        : 'I am not married'
+      }
+    </p>
+  </section>
+);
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+          {`I am ${person.age}`}
+        </p>
+      )
+    }
+
+    <p className="Person__partner">
+      {person.isMarried === true
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+        : 'I am not married'
+      }
```
etojeDenys, line 21:
```suggestion
      {person.isMarried
        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
        : 'I am not married'
      }
```
you can move some part of this logic to some var, it will improve the readability

src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age !== undefined
```
etojeDenys, line 9:
```suggestion
    {person.age
```
we can simplify it

etojeDenys (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
```
Viktor-Kost, line 4: it's a good practice to use object destructing 
```suggestion
export const Person = ({ person }) => {
  const {...} = person;
  
  return (
    <section className="Person">
```

TVChulkova: thx for the advice and review

PR #1935, opened 2023-12-09
add task solution

ApheliosGM:
[DEMO LINK](https://ApheliosGM.github.io/react_person/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,20 +1,3 @@
 iframe {
   display: none;
 }
-
-.Person {
-  width: fit-content;
-  margin-bottom: 16px;
-  padding: 8px;
-  border: 1px solid #000;
-  border-radius: 8px;
-
-  &__name {
-    margin-top: 8px;
-  }
-
-  &__age {
-    color: #00f;
-    font-weight: bold;
-  }
-}
--- a/src/App.scss
+++ b/src/App.scss
@@ -17,8 +17,4 @@ iframe {
     color: #00f;
     font-weight: bold;
   }
-
-  &__partner {
-    color: red;
-  }
 }
--- a/src/App.scss
+++ b/src/App.scss
@@ -17,4 +17,8 @@ iframe {
     color: #00f;
     font-weight: bold;
   }
+
+  &__partner {
+    color: red;
+  }
 }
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,25 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = sex => (
+    sex === 'm' ? 'wife' : 'husband'
+  );
```
nazarmatsevych, line 4:
```suggestion
  const partner = sex === 'm' ? 'wife' : 'husband'
```

No need to create function in this case

nazarmatsevych (changes requested):
Almost done, please check comment below

src/components/Person/Person.jsx
ApheliosGM: > No need to create function in this case

Incorrectly functions refuse to work properly.

etojeDenys: Nazar suggested that you can rewrite partner and use it as a variable instead of the function

etojeDenys (changes requested):

src/App.scss
```diff
@@ -17,4 +17,8 @@ iframe {
     color: #00f;
     font-weight: bold;
   }
+
+  &__partner {
+    color: red;
+  }
```
loralevitska, line 23:
```suggestion
```

loralevitska (changes requested):
Almost perfect) Check my comments for improvement

- try to pass all tests 
image
- marked text should be black
image

anastasiiavorobiova (changes requested):
Make sure to deploy your last fixes

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      <p className="Person__partner">
+        {
+          isMarried
+            ? `${partnerName} is my ${sex === 'm' ? `wife` : `husband`}`
+            : `I am not married`
+        }
```
etojeDenys, line 15: decide before jsx what text will be rendered and just render in jsx. something like that

```
```

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1942, opened 2023-12-13
implemented solution

GlibVK:
[DEMO LINK](https://glibvk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,30 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let partner = 'wife';
+
+  if (person.sex === 'f') {
+    partner = 'husband';
+  }
+
+  return (
+    <>
+      <section className="Person">
+        <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+        {person.age && (
+        <p className="Person__age">
+          {`I am ${person.age}`}
+        </p>
+        )}
+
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let partner = 'wife';
+
+  if (person.sex === 'f') {
+    partner = 'husband';
```
witflash, line 5: According to React logic, it's a bad idea to reassign some variables outside of the state.
Of course, you don't need state here for this static value, but the best way to do what you need is:
```
const partner = person.sex === 'f' ? 'husband' : 'wife';
```
Without any `if`

witflash (changes requested):
Good work!

Just one small comment. Check it, please, and make this change.

src/components/Person/Person.jsx
GlibVK: Thanks for advice. :)

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = (person.sex === 'f') ? 'husband' : 'wife';
+
+  return (
+    <>
```
etojeDenys, line 5: fragment is redundant here

etojeDenys (approved):
good job

PR #1950, opened 2023-12-17
add solution

Diliradon:
[CODE STYLE] - Use destructuring for getting access to values of props object.
[CODE STYLE] - Readabily is everything. Format ternary operator operands correctly - move each operand to the separate line.
[CODE STYLE] - Avoid putting several cases to conditional rendering. Create separate variable for the condition.


lerastarynets (approved):

PR #1956, opened 2023-12-21
Develop

DaniilBarilotti:
https://DaniilBarilotti.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,28 @@
-// export const Person = ({ person }) => ();
+const man = 'm';
+const woman = 'f';
+
+export const Person = ({
+  person: { name, age, sex, partnerName, isMarried },
+}) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+          ? <span>{`${partnerName} is my wife`}</span>
+          : null}
+
+        {isMarried && sex === woman
+          ? <span>{`${partnerName} is my husband`}</span>
+          : null}
+
+        {!isMarried
+          ? 'I am not married'
+          : null}
```
etojeDenys, line 24: try to simplify it, you can move some part of the logic to var

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
-// export const Person = ({ person }) => ();
+const man = 'm';
+const woman = 'f';
```
etojeDenys, line 2:
```suggestion
const MAN = 'm';
const WOMAN = 'f';
```
we usually name constants in uppercase

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #1973, opened 2024-01-02
Solution

VasiliyYarosh:


Xilston (approved):
Brilliant! Good job.

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const nameContent = `My name is ${person.name}`;
```
olya-shyian, line 2: you can use destructuring here

get properties from person like:
```
const {
  firstProperty,
  secondProperty,
  thirdProperty
} = person ; 
```

olya-shyian, line 2: there is no need to create const for ```const nameContent = `My name is ${person.name}`;```
...you can write it directly in the `return ()`

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+
+  let partnerContent;
+
+  if (person.isMarried) {
+    partnerContent = person.sex === 'f'
+      ? `${person.partnerName} is my husband`
+      : `${person.partnerName} is my wife`;
+  } else {
+    partnerContent = 'I am not married';
+  }
```
olya-shyian, line 12: don't complicate your code

you can just do this check:
const partner = sex === 'm' ? 'wife' : 'husband';

and then in the `return` write this condition:

```
  {isMarried
    ? `${partnerName} is my ${partner}`
     : 'I am not married'
    }
</p>
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+
+  let ageContent;
+
+  if (person.age !== undefined) {
+    ageContent = <p className="Person__age">{`I am ${person.age}`}</p>;
+  } else {
+    ageContent = null;
+  }
```
olya-shyian, line 20: here the same, don't create variable for this, just write it in the `return` using `&&` like
```{age && (```

**Review by olya-shyian — CHANGES_REQUESTED** [review]
Good job! Let's do some improvements🍀

**Comment by VasiliyYarosh** [comment]
DEMO  https://vasiliyyarosh.github.io/react_person/

File: src/components/Person/Person.jsx (hunk 3 above)
Comment (VasiliyYarosh) [reply src/components/Person/Person.jsx]: Thanks for the feedback! It helped me catch the points I didn't understand before.

File: src/App.jsx [hunk 4]
```diff
@@ -25,21 +26,8 @@ export const alex = {
...
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person person={misha} />
+    <Person person={olya} />
+    <Person person={alex} />
```
YegorKorenkov, line 31: You can use the same format style for jsx like for html(empty line between sibilings in this case)
```suggestion
    <Person person={misha} />
    
    <Person person={olya} />
    
    <Person person={alex} />
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+      {isMarried === true ? (
+        <p className="Person__partner">
+          {`${partnerName} is my ${partner}`}
+        </p>
+      ) : (
+        <p className="Person__partner">
+          I am not married
+        </p>
+      )
+      }
```
YegorKorenkov, line 27: You can change only content instead of re-rendering different paragraphs
```suggestion
      <p className="Person__partner">
        {isMarried ? (
          `${partnerName} is my ${partner'}`
        ) : (
          'I am not married'
        )}
      </p>
```

YegorKorenkov (approved):
Good job! But please pay attention on comments!

PR #1991, opened 2024-01-03
add task solution

sasholk:
[DEMO LINK](https://sasholk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.sex !== 'f' && (
+      <p className="Person__age">I am 37</p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${person.sex === 'f'
+          ? 'husband'
+          : 'wife'}`}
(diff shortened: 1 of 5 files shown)


MaxKalalashnyk02 (approved):
Great work!

src/components/Person/Person.jsx
```diff
@@ -1 +1,37 @@
...
+    isMarried,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age !== undefined && (
```
olya-shyian, line 18: you can just write simplier like ```{age && (```
because this value is automatically converted to boolean and if it is true than the next condition will work

src/components/Person/Person.jsx
```diff
@@ -1 +1,37 @@
...
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {`${partnerName} is my ${sex === 'f'
+              ? 'husband'
+              : 'wife'}`}
+          </p>
+        )
+        : (
+          <p className="Person__partner">I am not married</p>
```
olya-shyian, line 33: firstly you can do this check inside `p` tag, so you will not write it twice and write less code

secondly, create a variable for this check:
`const partner = sex === 'f' ? 'husband'  : 'wife'`

olya-shyian (changes requested):
nice work!🌷

src/components/Person/Person.jsx
sasholk: thx that looking better 😋

PR #1994, opened 2024-01-03
Done!(try1)

EdenParadise:
https://EdenParadise.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,37 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const {
+    name,
+    sex,
+    age,
+    partnerName,
+    isMarried,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age !== undefined && (
+        <p className="Person__age">
+          {`I am ${age}`}
+        </p>
+      )}
+
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {`${partnerName} is my ${sex === 'f'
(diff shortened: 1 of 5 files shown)


dyarechuk (approved):
Good Job!🔥
The only note is that the link to github pages doesn't open the page, did you use npm run deploy at the end?

src/components/Person/Person.jsx
```diff
@@ -1 +1,37 @@
...
+    isMarried,
+  } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">
+        {`My name is ${name}`}
+      </h2>
+
+      {age !== undefined && (
```
olya-shyian, line 18: there is no need to write this check for `undefined `.
you can write just `{age && (` because this variable will automatically be transformed to a boolean value...so if its true than part after `&&` will work

src/components/Person/Person.jsx
```diff
@@ -1 +1,37 @@
...
+          {`I am ${age}`}
+        </p>
+      )}
+
+      {isMarried
+        ? (
+          <p className="Person__partner">
+            {`${partnerName} is my ${sex === 'f'
+              ? 'husband'
+              : 'wife'}`}
```
olya-shyian, line 29: keep this condition in variable like
```const partner = sex === 'f' ? 'husband' : 'wife';``` and use it here

File: src/components/Person/Person.jsx [hunk 3]
```diff
@@ -1 +1,37 @@
...
+          </p>
+        )
+        : (
+          <p className="Person__partner">I am not married</p>
```
olya-shyian, line 33: also in order to write less code...you can write `p` tag only one time and use ternary check inside it

olya-shyian (changes requested):
Great work!🙂

PR #1996, opened 2024-01-04
add reat person solution

MaxKalalashnyk02:
[DEMO LINK](https://MaxKalalashnyk02.github.io/react_person/)

Implement a `Person` component rendering a person details using the given markup
and use it 3 times inside the `App` instead of static markup.

- if a person is not marriend - print `I am not married` message;
- keep the same classNames in `Person.jsx` as in `App.jsx` (`Person`, `Person__name`, `Person__age`, `Person__partner`).

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.sex !== 'f' && (
+      <p className="Person__age">I am 37</p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${person.sex === 'f' ? 'husband' : 'wife'}`}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
+  </section>
+);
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => (
```
olya-shyian, line 3: please use destructuring for person
```
const {
  firstProperty,
  secondProperty,
  thirdProperty
} = person;
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.sex !== 'f' && (
```
olya-shyian, line 9: I guess you have to check if `age` exists then ` && (`

also you wrote just hardcoded value `I am 37`, use `age`  variable here

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+      <p className="Person__age">I am 37</p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${person.sex === 'f' ? 'husband' : 'wife'}`}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
```
olya-shyian, line 18: keep this condition in variable like
```const partner = sex === 'f' ? 'husband' : 'wife';``` and use it here

also in order to write less code...you can write p tag only one time and use ternary check inside it

olya-shyian (changes requested):
Great!💫

SerhiiKirik (approved):
Well done 🔥

PR #1997, opened 2024-01-04
add task solution

Vladyslav-amp:
[DEMO LINK](https://Vladyslav-amp.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age ? <p className="Person__age">{`I am ${person.age}`}</p> : null}
+
+    <p className="Person__partner">
+      {person.isMarried === true ? `${person.partnerName} is my ${person.sex === 'f' ? `husband` : `wife`}` : `I am not married`}
+    </p>
+  </section>
+);
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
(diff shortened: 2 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age ? <p className="Person__age">{`I am ${person.age}`}</p> : null}
```
etojeDenys, line 7: we can simplify it using the && instead of the ?:

src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+
+    <p className="Person__partner">
+      {person.isMarried === true ? `${person.partnerName} is my ${person.sex === 'f' ? `husband` : `wife`}` : `I am not married`}
```
etojeDenys, line 10: decide before jsx what text will be rendered and just render in jsx.

etojeDenys (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
```
Viktor-Kost, line 2: it's a good practice to use object destructing
```suggestion
export const Person = ({ person }) => {
  const {...} = person;
  
  return (
    <section className="Person">
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
...
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
```
Viktor-Kost, line 4: no needs to use template string
```suggestion
      My name is {person.name}
```

Viktor-Kost (approved):

PR #2006, opened 2024-01-09
add person component

ly-manka:
[DEMO LINK](https://ly-manka.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+      {`My name is ${person.name}`}
+    </h2>
+
+    {person.age && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+
+    {person.isMarried
```
etojeDenys, line 13: you can add this check inside the `p` tag to simplify the "else" section

etojeDenys: decide before jsx what text will be rendered and just render in jsx. something like that
```
```

etojeDenys (changes requested):

nazarmatsevych (approved):
Awesome 🔥

PR #2021, opened 2024-01-18
solution

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+    {person.age !== undefined && (
+      <p className="Person__age">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+    {person.age !== undefined && (
+      <p className="Person__age">
+        {`I am ${person.age}`}
+      </p>
+    )}
+    <p className="Person__partner">
+
+      {person.isMarried === false ? (
+        `I am not married`
+      ) : `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`}
```
witflash, line 15: You can move the second ternary operator to the separate constant:
```
const partnerStatus = person.sex === 'm' ? 'wife' : 'husband';
```
So then you can use simpler syntax:
```
: `${person.partnerName is my ${partnerStatus}`
```
It looks better

witflash (approved):
Well done 👍🏻 
Check my comment for your future improvements

PR #2027, opened 2024-01-20
Solution

ghost:


src/App.jsx
```diff
@@ -24,22 +25,17 @@ export const alex = {
...
-  <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
+  <>
+    <Person
+      person={misha}
+    />
```
anastasiia-tilikina, line 31:
```suggestion
    <Person person={misha} />
```
we don't need to write props from new line if there is only one check all cases

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+    }
+
+    <p className="Person__partner">
+      {person.isMarried ? (
+        `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+      ) : (
+        'I am not married'
+      )
+      }
+    </p>
```
anastasiia-tilikina, line 21: do not make nested ternary operators, move one condition into fn or variable

anastasiia-tilikina (changes requested):
GJ!

just a couple issues

src/components/Person/Person.jsx
etojeDenys: image
still not fixed


decide before jsx what text will be rendered and just render in jsx.

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #2040, opened 2024-01-24
solution

yurovych:
https://yurovych.github.io/react_person/


src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+
+      {person.isMarried ? (
+        <p className="Person__partner">
+          {`${person.partnerName} is my ${who}`}
+        </p>
+      ) : (
+        <p className="Person__partner">
+          I am not married
+        </p>
+      )}
```
etojeDenys, line 24: you can add this condition only for the text in p tag

etojeDenys (approved):
good job

PR #2046, opened 2024-01-25
person

NikolaienkoN:
https://NikolaienkoN.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">
+      {`My name is ${person.name}`}
+    </h2>
+    {person.age && (
+      <p className="Person__age">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+      </p>
+    ) : (
+      <p className="Person__partner">
+        {`${person.partnerName} is my wife`}
+      </p>
+    )}
+    {!person.isMarried && (
+      <p className="Person__partner">
+        I am not married
+      </p>
```
etojeDenys, line 23: image
looks broken


```
```

etojeDenys (changes requested):

src/components/Person/Person.jsx
maxim2310: not fixed
image

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

nazarmatsevych (approved):
👍

PR #2056, opened 2024-01-29
add task solution

andreyysak:
[DEMO LINK](https://andreyysak.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,35 +1,24 @@
 import './Person.scss';
 
-export const Person = ({ person }) => (
+export const Person = ({ person }) => {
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,35 @@
...
+            </>
+          )
+          }
+        </>
+      ) : (
+        <>
+          I`m not married
+        </>
+      )}
+    </p>
```
etojeDenys, line 33: decide before jsx what text will be rendered and just render in jsx. something like that

```
```

etojeDenys (changes requested):

lerastarynets (approved):
lgtm 🚀

PR #2070, opened 2024-02-05
add task solution

ArsenSkandalenko:


witflash (changes requested):
Hi!

There is a small problem with your task. Maybe you've been a little confused by this task, but:
- you shouldn't need to change the file `Person.spec.jsx`. This file is just for tests. You need to leave it as is. 
- you have to modify `App.jsx` component to render your `Person` component. Something like this:
```
export const App = () => (
  <div className="App">
    <Person person={misha} />
    <Person person={olya} />
    <Person person={alex} />
  </div>
);
```

Try to fix these problems, then commit, push, and request review again.

src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+
+        {person.name}
+      </h2>
+      { person.age ? (
+        <p className="Person__age">
+          {`I am `}
+
+          {person.age}
+        </p>
+      ) : ''}
```
etojeDenys, line 19:
```suggestion
      { person.age && (
        <p className="Person__age">
          {`I am ${person.age}`} 
        </p>
```

src/components/Person/Person.spec.jsx
```diff
@@ -1,73 +1,47 @@
...
+  partnerName: 'Natasha',
+};
+export const olya = {
+  name: 'Olya',
+  sex: 'f',
+  isMarried: true,
+  partnerName: 'Maksym',
+};
+export const alex = {
+  name: 'Alex',
```
etojeDenys, line 19: why do you change the tests?

etojeDenys (changes requested):

etojeDenys (changes requested):
you don't push changes, don't hesitate to ask for some help with that in the chat

PR #2097, opened 2024-02-17
added solution react person

nk8bnj:
[DEMO LINK](https://nk8bnj.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const married = (sex, partnerName) => {
```
BudnikOleksii, line 2: - You can move this function outside the component
- Also, the function name should have a verb, and describe what it does, for example, `getPartnerText`

BudnikOleksii (approved):
👍

PR #2101, opened 2024-02-19
resolve task

zozulia-developer:
[DEMO LINK](https://zozulia-developer.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,11 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
Viktor-Kost, line 1: it's a good practice to use object destructing

Viktor-Kost (approved):
GJ!

PR #2111, opened 2024-02-23
Add solution

ViktoriiaRepo:
[DEMO LINK](https://viktoriiarepo.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,16 +1,7 @@
 export const Person = ({ person }) => {
   const { name, age, sex, isMarried, partnerName } = person;
   const isMale = sex === 'm';
-  const isMarriedMale = isMale && isMarried;
-  const isMarriedFemale = !isMale && isMarried;
-
-  let partnerTitle;
-
-  if (isMarriedMale || isMarriedFemale) {
-    partnerTitle = isMale ? 'wife' : 'husband';
-  } else {
-    partnerTitle = 'not married';
-  }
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const isMale = sex === 'm';
+  const isMarriedMale = isMale && isMarried;
+  const isMarriedFemale = !isMale && isMarried;
```
BudnikOleksii, line 5: These variables are redundant as all what you actually need is `isMarried`

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
...
+
+  let partnerTitle;
+
+  if (isMarriedMale || isMarriedFemale) {
+    partnerTitle = isMale ? 'wife' : 'husband';
+  } else {
+    partnerTitle = 'not married';
+  }
```
BudnikOleksii, line 13: - `partnerTitle = 'not married';` it doesn't make sense as you don't show a parent title if a person isn't married
- could be simplified just with `const partnerTitle = isMale ? 'wife' : 'husband';`

BudnikOleksii (changes requested):
Good work 👍 
Let's clean your solution slightly

BudnikOleksii (approved):
Good job👍

PR #2114, opened 2024-02-24
solution

robsza1122:


src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const isFemale = sex === 'f';
+  const isMale = sex === 'm';
```
BogdanMaliuta, line 4: dont need to create 2 variables, just do your conditions !isFemale or !isMale

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${nameOfPartner}`
+          : 'I am not married'}
+      </p>
```
BogdanMaliuta, line 24: add empty lines between siblings, like in HTML

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```
BogdanMaliuta, line 2: when your destruction haves more than 2 args split it in rows

BogdanMaliuta (changes requested):
GJ!
Need code style fix

src/components/Person/Person.jsx
robsza1122: Lint says that is ok. Should I ignore it?

yevhenii-pyl: It mostly depends on a project, as a rule of thumb - whenever a line goes too long we should split it in rows, it makes even more sense when you want to extract properties on the props level: 

```
export const Person = ({ 
    person: {
         name,...
    } 
}) => {
```

It just makes your code more readable and it's easier to find / modify something in props. 
Even if linter still don't mark it as an error, it still makes sense to write arguments in a row.

It's not a mistake here in terms of this assignment, but in 90% your reviewer at work will ask you for these changes, so it's a good idea to develop this code-styling behaviour.

src/components/Person/Person.jsx
```diff
@@ -1 +1,29 @@
...
+  const isMale = sex === 'm';
+  const marriedFemale = !isMale && isMarried;
+  const marriedMale = isMale && isMarried;
+
+  let nameOfPartner;
```
yevhenii-pyl, line 7: Two things here:

1. You modify this variable into a string later, so it's better to declare it as an empty string, to somehow declare a future type of value that will be placed here. 
2.  Not exactly the best naming here. `nameOfPartner` ? Maybe, `partnerStatus` or something like that will do better, because when you see 'name' you probably expect to receive a name here, like 'Magda'.

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
-// export const Person = ({ person }) => ();
+export const Person = ({
+  person: { name, age, sex, isMarried, partnerName },
```
BogdanMaliuta, line 2: split it into lines

BogdanMaliuta (approved):
GJ!
Check my comment and fix it

PR #2131, opened 2024-03-02
solution

Arkadiuszx:
[DEMO LINK](https://Arkadiuszx.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,45 +1,12 @@
 import React from 'react';
 import './App.scss';
-
-export const misha = {
-  name: 'Misha',
-  age: 37,
-  sex: 'm',
-  isMarried: true,
-  partnerName: 'Natasha',
-};
-
-export const olya = {
-  name: 'Olya',
-  sex: 'f',
-  isMarried: true,
-  partnerName: 'Maksym',
-};
-
-export const alex = {
-  name: 'Alex',
-  age: 25,
-  sex: 'm',
-  isMarried: false,
-};
+import { Person } from './components/Person/Person';
+import { people } from './components/People/People';
 
 export const App = () => (
   <div className="App">
-    <section className="Person">
-      <h2 className="Person__name">My name is Misha</h2>
-      <p className="Person__age">I am 37</p>
-      <p className="Person__partner">Natasha is my wife</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Olya</h2>
-      <p className="Person__partner">Maksym is my husband</p>
-    </section>
-
-    <section className="Person">
-      <h2 className="Person__name">My name is Alex</h2>
-      <p className="Person__age">I am 25</p>
-      <p className="Person__partner">I am not married</p>
-    </section>
+    <Person person={people.misha} />
+    <Person person={people.olya} />
+    <Person person={people.alex} />
   </div>
 );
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -1,45 +1,12 @@
...
-};
-
-export const alex = {
-  name: 'Alex',
-  age: 25,
-  sex: 'm',
-  isMarried: false,
-};
+import { Person } from './components/Person/Person';
+import { people } from './components/People/People';
```
yevhenii-pyl, line 4: It is definitely NOT a component :) 

If you want to separate people values from your component => you can create a simple helpers.js file near this component, so the structure will be:

components
   // => People
        // => People.js
        // => helpers.js
        // => index.js (imports People component and exports it: `export { default } from './People';`)

src/App.jsx
```diff
@@ -1,45 +1,12 @@
...
-  partnerName: 'Maksym',
-};
-
-export const alex = {
-  name: 'Alex',
-  age: 25,
-  sex: 'm',
-  isMarried: false,
-};
+import { Person } from './components/Person/Person';
```
yevhenii-pyl, line 3: If you follow my previous comment and will add index.js to component directory - this will simplify import here, it will become just: `import { Person } from './components/Person';` since index.js is always considered as a default directory for importing something from the folder.

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  let partner;
+
+  if (isMarried) {
```
yevhenii-pyl, line 5: This should also be extracted into a separate helpers file as a helper function, like `generatePatnerDescription(person)`.

Could you also think how to simplify this function?

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```
GJ, props should also be destructed here or at the highest level.

yevhenii-pyl (changes requested):
Good job!
Let's adjust a few things and we are good to go.

Arkadiuszx:
Is this how it's supposed to look?

yevhenii-pyl (changes requested):
Solid work, you got this. 
Let's just fix that minor thing with people object to make it right.

src/helpers.js
```diff
@@ -0,0 +1,21 @@
+export const People = {
```
yevhenii-pyl, line 1: Yep, but it should be lowercased :) 

In react, you use camelCase with uppercased first letter only for components.

yevhenii-pyl (approved):
You got the idea, thus there is no reason to block the PR any further.
HOWEVER, please, make sure you are following naming conventions. 

**components/variables:** 
component => CamelCase => `PeopleTable` 
variables => camelCase => `pepleArray`

**files:** 
Generally, there are no direct rules, you can just follow best practices and common logic. Exact rules will be stated by your team lead at each project. 
However, consistency is always the key here, if you put 'any-helpers' at files named `helpers` => you should always call that type of files only `helpers`.
There is no need to call one file `helpers` and second `helperFunctions` => this is just unnecessary. 

Looking at your `components/Person` folder you can already imagine the right structure for basically ANY component folder you will create:

```
=> component folder 
   => Component.js
   => index.js
   => helpers.js 
   => ADDITIONALY: here may be file for component specific styles
   => ADDITIONALY: LittleComponent.js (the component that is basically just a helper here and is directly attached to the main component but you just want to have them in separate files, this one is never exported outside the folder and used only by the main component here)
```

PR #2134, opened 2024-03-02
react_person

p1zzat1me:
https://p1zzat1me.github.io/react_person/


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+      {person.isMarried ? (
+        <p className="Person__partner">
+          {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
etojeDenys, line 12: decide before jsx what text will be rendered and just render in jsx. something like that

```
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <>
```
etojeDenys, line 2: fragment is redundant here

https://react.dev/reference/react/Fragment#usage

etojeDenys (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #2136, opened 2024-03-03
Person compose data

DenNice-cloud:
[DEMO LINK](https://DenNice-cloud.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age !== undefined ? (
+      <p className="Person__age">I am {person.age}</p>
+    ) : null}
+
+    <p className="Person__partner">
+      {person.isMarried === false
+        ? 'I am not married'
+        : `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`}
+    </p>
+  </section>
+);
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,7 @@
+/* eslint-disable prettier/prettier */
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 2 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age !== undefined ? (
+      <p className="Person__age">I am {person.age}</p>
+    ) : null}
```
olduma, line 7: `{age && <p className="Person__age">I am {person.age}</p>}`

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
olduma, line 1: I think it's easier to destruct **person** here to avoid confusion and write less code!

olduma (approved):
Good job!

src/components/Person/Person.jsx
BudnikOleksii: > I think it's easier to destruct **person** here to avoid confusion and write less code!

Make sense, but it's not so important in this case, you can choose

src/components/Person/Person.jsx
BudnikOleksii: > `{person.age && <p className="Person__age">I am {person.age}</p>}`

Good point, it doesn't make sense to compare with `undefined`. And I don't see any reason for `null`

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age !== undefined ? (
+      <p className="Person__age">I am {person.age}</p>
+    ) : null}
+
+    <p className="Person__partner">
+      {person.isMarried === false
```
BudnikOleksii, line 10: The same, don't compare with boolean variables
https://developer.mozilla.org/en-US/docs/Glossary/Falsy
https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+        ? 'I am not married'
+        : `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`}
```
BudnikOleksii, line 12: Ternary operator inside ternary operator is a bad practice, create a variable for this value to avoid it

BudnikOleksii (approved):
Almost done, let's improve your solution

PR #2145, opened 2024-03-04
react_person. Solution

ivan-gritsenko:
[DEMO LINK](https://ivan-gritsenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = { ...person };
+  let partnerStatus;
+
+  if (isMarried) {
+    partnerStatus =
+      sex === 'm'
+        ? `${partnerName} is my wife`
+        : `${partnerName} is my husband`;
+  } else {
+    partnerStatus = 'I am not married';
+  }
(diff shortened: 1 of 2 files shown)


Graf6612 (approved):
Hello,Good job))

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = { ...person };
+  let partnerStatus;
+
+  if (isMarried) {
+    partnerStatus =
+      sex === 'm'
+        ? `${partnerName} is my wife`
+        : `${partnerName} is my husband`;
```
BudnikOleksii, line 9: Only the word wife/husband is different, so set it conditionally instead of repeating all

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+  } else {
+    partnerStatus = 'I am not married';
+  }
```
BudnikOleksii, line 12: You can set it by default and get rid of `else` statement

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution a bit

BudnikOleksii (approved):
👍

PR #2146, opened 2024-03-04
Add sol

Graf6612:


vlesia (approved):
Great job! :fire:

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
```
Moroz-Dmytro, line 3: image

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    <p className="Person__partner">{isMarried(person)}</p>
+  </section>
+);
+
+function isMarried(person) {
```
Moroz-Dmytro, line 9: Better to create variable for this, not function

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+  return person.isMarried
+    ? `${person.partnerName} is my ${sex(person)}`
+    : 'I am not married';
+}
+
+function sex(person) {
```
Moroz-Dmytro, line 15: Also, make this as a variable

Moroz-Dmytro (changes requested):
Good progress 👍 
But need to make some fixes
Also, pay attention to checklist, before sending to review

src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  let partner = 'I am not married';
```
BudnikOleksii, line 3: Variable initialization is redundant here as you return the same string from the `determinePartnerStatus` function

BudnikOleksii (approved):
👍

PR #2152, opened 2024-03-04
add task solution

Math8585:
- [DEMO LINK](https://Math8585.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  let marriedStatus = '';
+
+  if (isMarried) {
+    marriedStatus =
+      sex === 'm'
+        ? `${partnerName} is my wife`
+        : `${partnerName} is my husband`;
+  } else {
+    marriedStatus = 'I am not married';
+  }
+
+  return (
+    <section className="Person">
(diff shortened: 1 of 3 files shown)


Dmitrel (approved):
Good job!

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+  let marriedStatus = '';
+
+  if (isMarried) {
+    marriedStatus =
+      sex === 'm'
+        ? `${partnerName} is my wife`
+        : `${partnerName} is my husband`;
+  } else {
+    marriedStatus = 'I am not married';
+  }
```
BudnikOleksii, line 12: Set it buy default and you'll get rid of `else` statement

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  let marriedStatus = '';
+
+  if (isMarried) {
+    marriedStatus =
+      sex === 'm'
+        ? `${partnerName} is my wife`
+        : `${partnerName} is my husband`;
```
BudnikOleksii, line 9: In both cases the string is almost the same, only the word wife/husband is different, so you can set only it conditionally

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

BudnikOleksii (approved):
👍

PR #2154, opened 2024-03-04
add solution

okarnaukh:
[DEMO LINK](https://okarnaukh.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,16 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+
+    {person.isMarried &&
+      (person.sex === 'm' ? (
+        <p className="Person__partner">{person.partnerName} is my wife</p>
+      ) : (
+        <p className="Person__partner">{person.partnerName} is my husband</p>
+      ))}
+
+    {!person.isMarried && <p className="Person__partner">I am not married</p>}
+  </section>
+);
--- a/README.md
+++ b/README.md
@@ -13,4 +13,4 @@ and use it 3 times inside the `App` instead of static markup.
 - Install Prettier Extention and use this [VSCode settings](https://mate-academy.github.io/fe-program/tools/vscode/settings.json) to enable format on save.
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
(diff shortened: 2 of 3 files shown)


Vlobdam (approved):
👍

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
...
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+
+    {person.isMarried &&
+      (person.sex === 'm' ? (
+        <p className="Person__partner">{person.partnerName} is my wife</p>
+      ) : (
+        <p className="Person__partner">{person.partnerName} is my husband</p>
+      ))}
+
+    {!person.isMarried && <p className="Person__partner">I am not married</p>}
```
BudnikOleksii, line 14: - Use the ternary operator when you have two options(person.isMarried) like in this case.
- <p className="Person__partner"> is the same in all cases, don't repeat it, only content is different 
- `{person.partnerName} is my ` here the content is almost the same, the only different is in word husband/wife, so you can set only it conditonally

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution a bit

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
```
Moroz-Dmytro, line 3: image

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+
+    <p className="Person__partner">
+      {person.isMarried ? (
+        <>
```
Moroz-Dmytro, line 9: You don't need fragment here

src/components/Person/Person.jsx
Moroz-Dmytro: Use destructuring for all fields

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+          {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
```
Moroz-Dmytro, line 10: Create variable for `person.sex === 'm' ? 'wife' : 'husband'`

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+        </>
+      ) : (
+        <>I am not married</>
```
Moroz-Dmytro, line 13: Useless fragment

Moroz-Dmytro (changes requested):
Good progress 👍 
But we can improve your solution

stassokolov1 (approved):
Looks good to me, I'm approving it🔥

PR #2156, opened 2024-03-04
add solution

Vlobdam:
[DEMO LINK](https://vlobdam.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,9 +1,7 @@
 import './Person.scss';
 
-const IsMarriedParagraph = (sex, partnerName) => {
-  return sex === 'f'
-    ? `${partnerName} is my husband`
-    : `${partnerName} is my wife`;
+const getMarriedInfo = (sex, partnerName) => {
+  return `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`;
 };
 
 export const Person = ({ person }) => {
@@ -14,7 +12,7 @@ export const Person = ({ person }) => {
       <h2 className="Person__name">My name is {name}</h2>
       {age && <p className="Person__age">I am {age}</p>}
       <p className="Person__partner">
-        {isMarried ? IsMarriedParagraph(sex, partnerName) : 'I am not married'}
+        {isMarried ? getMarriedInfo(sex, partnerName) : 'I am not married'}
       </p>
     </section>
   );
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
-import './App.scss';
+
 import { Person } from './components/Person/Person';
+
+import './App.scss';
 
 export const misha = {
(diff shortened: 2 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+/* eslint-disable no-nested-ternary */
+import './Person.scss';
+
+export const Person = ({ person }) => (
```
hattoriKanto, line 4: Use destructuring for getting access to values of `props` object.

BAD EXAMPLE:
```jsx
export const Component = props => (
  <>
    <h1>{props.firstProperty}</h1>
    <span>{props.secondProperty}</span>
    <p>{props.thirdProperty}</p>
  </>
);
```

GOOD EXAMPLE:
```jsx
export const Component = ({
 firstProperty,
 secondProperty,
 thirdProperty
}) => (
  <>
    <h1>{firstProperty}</h1>
    <span>{secondPropery}</span>
    <p>{thirdProperty}</p>
  </>
);
```

ALSO GOOD EXAMPLE: 

```jsx
export const Component = (props) => {
 const {
  firstProperty,
  secondProperty,
  thirdProperty
} = props; 

 return (
  <>
    <h1>{firstProperty}</h1>
    <span>{secondPropery}</span>
    <p>{thirdProperty}</p>
  </>
 )
};
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
...
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    <p className="Person__partner">
+      {person.isMarried
+        ? person.sex === 'f'
+          ? `${person.partnerName} is my husband`
+          : `${person.partnerName} is my wife`
```
hattoriKanto, line 12: Try to avoid nested ternary operators because it's reduce code readable. For example, you can create a variable and assign ternary statement to it. As for me, not a big mistake, so if you want you can just ignore it))

[ESLint](https://eslint.org/docs/latest/rules/no-nested-ternary#resources)

hattoriKanto (changes requested):
Looks good but need some fixes to be implemented.

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+
+export const Person = ({ person }) => {
+  const { name, age, isMarried, sex, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      <p className="Person__partner">
+        {isMarried ? getMarriedInfo(sex, partnerName) : 'I am not married'}
```
Moroz-Dmytro, line 15: But I would recommend not to use function, better to make it as variable

PR #2159, opened 2024-03-04
add task solution

hattoriKanto:
[DEMO LINK](https://hattoriKanto.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const SPOUSE = sex === 'f' ? 'husband' : 'wife';
+  const NOT_MARRIED_MESSAGE = 'I am not married';
```
BudnikOleksii, line 6: - The "SPOUSE" variable is not constant as it is set dynamically, so they should be in camelCase
- NOT_MARRIED_MESSAGE constant you can move outside the component

BudnikOleksii (approved):
Good work 👍 
Approved, but left one note about constants

PR #2169, opened 2024-03-06
add task solution

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,17 +1,18 @@
 export const Person = ({ person }) => {
   const { name, age, sex, isMarried, partnerName } = person;
+  const partnerGender = sex === 'm' ? 'wife' : 'husband';
 
   return (
     <section className="Person">
       <h2 className="Person__name">My name is {name}</h2>
       {age && <p className="Person__age">I am {age}</p>}
-      {!isMarried && <p className="Person__partner">I am not married</p>}
-      {isMarried && sex === 'm' && (
-        <p className="Person__partner">{partnerName} is my wife</p>
-      )}
-      {isMarried && sex === 'f' && (
-        <p className="Person__partner">{partnerName} is my husband</p>
-      )}
+      {
+        <p className="Person__partner">
+          {isMarried
+            ? `${partnerName} is my ${partnerGender}`
+            : 'I am not married'}
+        </p>
+      }
     </section>
   );
 };


src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let contentPartner;
+  let contentAge;
+
+  if ('age' in person) {
```
yevhenii-pyl, line 5: Nope, definitely not like that. You never want to assign an html-like code to a variable inside the react component in that manner. 

If you wish to extract the logic with age and partner - there should be a helper function for it, like:
```

const renderAge = (age) => {
  if(person?.age) {
    return <p className="Person__age">I am {person.age}</p>;
  }
  return null;
}

...
return (
    <section className="Person">
      <h2 className="Person__name">My name is {person.name}</h2>
      {renderAge()}
    </section>
)
```

However, this approach may also be simplified in this case by just: 

```
  return (
    <section className="Person">
      <h2 className="Person__name">My name is {person.name}</h2>
      {person.age && <p className="Person__age">I am {person.age}</p>}
    </section>
  );
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
...
+    contentAge = <p className="Person__age">I am {person.age}</p>;
+  } else {
+    contentAge = null;
+  }
+
+  if (person.isMarried === false) {
```
yevhenii-pyl, line 11: Same here, it's a no go

src/components/Person/Person.jsx
```diff
@@ -1 +1,30 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
```
yevhenii-pyl, line 1: Let's extract properties from the person here: 

`({ person: { age, name ....} })`

yevhenii-pyl (changes requested):
Nice work, but it requires some refactoring :)

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      {!isMarried && <p className="Person__partner">I am not married</p>}
+      {isMarried && sex === 'm' && (
+        <p className="Person__partner">{partnerName} is my wife</p>
+      )}
+      {isMarried && sex === 'f' && (
```
yevhenii-pyl, line 12: Let's simplify this one even more:
create a constant for a partnerGender -   `const partnerGender = sex === 'm' ? 'wife' : 'husband'`

Modify jsx:
```
        {isMarried
          ? `${partnerName} is my ${partnerGender}`
          : 'I am not married'}
      </p>
```

yevhenii-pyl (changes requested):
Just one more simplification and we are done here ;)

yevhenii-pyl (approved):
🚀

PR #2173, opened 2024-03-07
solution

Radymyr:
[DEMO LINK](https://Radymyr.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner">i am not married</p>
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
```
nazarmatsevych, line 4: You can destructure person object

src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
```
nazarmatsevych, line 7: Nested ternary looks complicated, let's create a separate variable for person partner and use it here

src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+      </p>
+    ) : (
+      <p className="Person__partner">i am not married</p>
```
nazarmatsevych, line 10: You can use just one `<p>` tag and change only it's content

nazarmatsevych (changes requested):
Good start, please fix the comments below

etojeDenys (approved):
awesome

PR #2179, opened 2024-03-08
add task solution

Marys888:
[DEMO LINK](https://Marys888.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, sex, isMarried, partnerName, age } = person;
+
+  return (
+    <section className="Person">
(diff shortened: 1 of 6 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  const { name, sex, isMarried, partnerName, age } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} ${sex === 'm' ? 'is my wife' : 'is my husband'}`
+          : 'I am not married'}
```
etojeDenys, line 11: nested ternary operators are a bad idea


```
```

etojeDenys (changes requested):

nazarmatsevych (approved):
Well done 👍

PR #2185, opened 2024-03-10
v: 1

Stan1slav123:
[DEMO LINK](https://Stn1slav123.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age !== undefined && (
+      <p className="Person__age">{`I am ${person.age}`}</p>
+    )}
+    <p className="Person__partner">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age !== undefined && (
```
etojeDenys, line 4:
```suggestion
    {person.age && (
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+      <p className="Person__age">{`I am ${person.age}`}</p>
+    )}
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+        : 'I am not married'}
```
etojeDenys, line 10: nested ternary operators are a bad practice



```
```

etojeDenys (changes requested):

PR #2198, opened 2024-03-16
done

VadymBryvka:
[DEMO](https://vadymbryvka.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,17 +1,21 @@
 // export const Person = ({ person }) => ();
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">My name is {person.name}</h2>
-    {person.age && <p className="Person__age">I am {person.age}</p>}
-    <p className="Person__partner">
-      {person.isMarried ? (
-        <>
-          {person.partnerName} is my{' '}
-          {person.sex === 'f' ? <>husband</> : <>wife</>}
-        </>
-      ) : (
-        <>I am not married</>
-      )}
-    </p>
-  </section>
-);
+
+export const Person = ({ person }) => {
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
 // export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    <p className="Person__partner">
+      {person.isMarried ? (
+        <>
+          {person.partnerName} is my{' '}
+          {person.sex === 'f' ? <>husband</> : <>wife</>}
```
nazarmatsevych, line 10: Nested ternary looks complicated, let's create a separate variable and use it here

`const partner = person.sex === 'f' ? 'husband' : 'wife'`

nazarmatsevych (changes requested):
Almost done, let's make some improvements

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+      <p className="Person__partner">
+        {person.isMarried ? (
+          <>
+            {person.partnerName} is my {partner}
+          </>
+        ) : (
+          <>I am not married</>
+        )}
```
VitaliyBondarenko1982, line 17:
```suggestion
        {person.isMarried 
          ?  `${person.partnerName} is my ${partner}`
          : 'I am not married'
        }
```
no need to wrap in fragment text. it will  be just string.
Also will be better move this ternary in component (below partner ternary) and render just final text in jsx.

VitaliyBondarenko1982 (approved):
Good job.

PR #2216, opened 2024-03-19
annihilated

true-binar-White-Rabbit:
[DEMO LINK](https://true-binar-White-Rabbit.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+      {person.isMarried === true ? (
+        <p className="Person__partner">
+          {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
etojeDenys, line 12: 1. you can add this condition to the text in the tag instead of the whole tag
2. nested ternary operators are a bad practice


```
```

etojeDenys (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,14 @@
...
+  const personGender = person.sex === 'm' ? 'wife' : 'husband';
+  const personStatus = person.isMarried
+    ? `${person.partnerName} is my ${personGender}`
+    : 'I am not married';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+      <p className="Person__partner">{personStatus}</p>
```
volodymyr-soltys97, line 11:
```suggestion
      <h2 className="Person__name">My name is {person.name}</h2>
      
      {person.age && <p className="Person__age">I am {person.age}</p>}
      
      <p className="Person__partner">{personStatus}</p>
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,14 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const personGender = person.sex === 'm' ? 'wife' : 'husband';
```
volodymyr-soltys97, line 2: Use destruction for `person`

volodymyr-soltys97 (changes requested):
Almost done!
Let's make your code better

src/components/Person/Person.jsx
true-binar-White-Rabbit: not so important in this case

etojeDenys (approved):

PR #2226, opened 2024-03-23
Solution of task react_person

Voloshynd:
[DEMO LINK](https://Voloshynd.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+import './Person.scss';
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    {person.isMarried ? (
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
```
etojeDenys, line 13: you can move the condition to the text inside the tag

and nested ternary operators are a bad practice

```
```

etojeDenys (changes requested):

src/components/Person/Person.jsx
Voloshynd: Thank's for feedback

etojeDenys (approved):
lgtm

PR #2235, opened 2024-03-28
Solution

opalahecha:
[DEMO LINK] - https://opalahecha.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      <p className="Person__partner">
+        {isMarried &&
+          (sex === 'f'
+            ? `${partnerName} is my husband`
+            : `${partnerName} is my wife`)}
```
lerastarynets, line 12: consider moving this expression to a separate const

lerastarynets (approved):
Nice start! Let's fix and improve your work a little bit)

src/components/Person/Person.jsx
opalahecha: Can I use // eslint-disable-next-line no-nested-ternary before const or will be better let and  if else? 
Because an error message is shown: Do not nest ternary expressions.

PR #2241, opened 2024-04-01
task Person solution

thisIvanRepo:
[DEMO LINK](https://IvanPovich.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,17 +1,17 @@
 export const Person = ({ person }) => {
   const { name, age, sex, isMarried, partnerName } = person;
 
+  const agePerson = age ? <p className="Person__age">I am {age}</p> : null;
+  const personPartnerType = sex === 'm' ? 'wife' : 'husband';
+  const personPartner = isMarried
+    ? `${partnerName} is my ${personPartnerType}`
+    : 'I am not married';
+
   return (
     <section className="Person">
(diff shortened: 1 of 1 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried ? (
+        <p className="Person__partner">
+          {partnerName} is my {sex === 'm' ? 'wife' : 'husband'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
etojeDenys, line 14: nested ternary operators are a bad practice

```
```

etojeDenys (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  const agePerson = age ? <p className="Person__age">I am {age}</p> : null;
```
etojeDenys, line 4: it's redundant to move this condition to the var

etojeDenys (approved):
good job

PR #2253, opened 2024-04-04
Develop

MOODDDII:
[Demo](https://moodddii.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
```
anastasiia-tilikina, line 1: You could use destructuring for person

anastasiia-tilikina (approved):
Well done!
just consider my comment for the future

PR #2256, opened 2024-04-05
Solution

Allanpt:
- [DEMO LINK](https://Allanpt.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -6,7 +6,7 @@ export const Person = ({ person }) => {
     <section className="Person">
       <h2 className="Person__name">My name is {person.name}</h2>
 
-      {person.age !== undefined && (
+      {person.age && (
         <p className="Person__age">I am {person.age}</p>
       )}
 


src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+
+      {person.age !== undefined && (
```
DarkMistyRoom, line 9:
```suggestion
      {person.age && (
```

DarkMistyRoom: It gives the same result because of type conversion

src/components/Person/Person.jsx
```diff
@@ -1 +1,24 @@
...
+        <p className="Person__age">I am {person.age}</p>
+      )}
+
+      {person.isMarried ? (
+        <p className="Person__partner">
+          {person.partnerName} is my {person.sex === 'f' ? 'husband' : 'wife'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
DarkMistyRoom, line 19: You can move the ternary operator to a variable to avoid duplication and increase readability. Also, it would be better to avoid ternary operators nesting

DarkMistyRoom (approved):
Good job! Take a look at the comments

PR #2257, opened 2024-04-07
VB Person solution

vbuldenko:
[DEMO LINK](https://vbuldenko.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+
+      {person.isMarried ? (
+        <p className="Person__partner">
+          {person.partnerName} is my {person.sex === 'f' ? 'husband' : 'wife'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
etojeDenys, line 14: nested ternary operators are a bad practice

```
```

etojeDenys (changes requested):

nazarmatsevych (approved):
👍

PR #2259, opened 2024-04-07
task solution

kinqbert:
[DEMO LINK](https://kinqbert.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let partnerString = '';
+
+  if (person.isMarried) {
+    if (person.sex === 'm') {
+      partnerString = `${person.partnerName} is my wife`;
+    } else {
+      partnerString = `${person.partnerName} is my husband`;
+    }
+  } else {
+    partnerString = 'I am not married';
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      {person.age !== undefined && (
+        <p className="Person__age">{`I am ${person.age}`}</p>
+      )}
+      <p className="Person__partner">{partnerString}</p>
+    </section>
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+      partnerString = `${person.partnerName} is my husband`;
+    }
+  } else {
+    partnerString = 'I am not married';
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+      {person.age !== undefined && (
```
DanilWeda, line 17:
```suggestion
      {person.age && (
```
The better way doesn't use type condition.

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+
+  if (person.isMarried) {
+    if (person.sex === 'm') {
+      partnerString = `${person.partnerName} is my wife`;
+    } else {
+      partnerString = `${person.partnerName} is my husband`;
+    }
+  } else {
+    partnerString = 'I am not married';
+  }
```
DanilWeda, line 12: You can write func that returns to you ready string

DanilWeda (changes requested):
Good job! 
Check small comments for improvement.
Thanks)

kinqbert:
Thank's four your review! Added mentioned improvements to the code 👍

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+  if (person.isMarried) {
+    if (person.sex === 'm') {
+      return `${person.partnerName} is my wife`;
+    }
+
+    return `${person.partnerName} is my husband`;
+  }
+
+  return 'I am not married';
+}
```
alexander-ignatow, line 21: Technically it is not an error, but for the readability, it is better define functions before using them

alexander-ignatow (approved):
Looks good! 👍

PR #2267, opened 2024-04-10
add_solution

metrakonon:
[DEMO](https://metrakonon.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+const genderTitle = {
+  m: 'wife',
+  f: 'husband',
+};
+
+export const Person = ({ person }) => {
+  const { name, age, partnerName, isMarried, sex } = person;
+  const genderLine = `${partnerName} is my ${genderTitle[sex]}`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried ? (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried ? (
+        <p className="Person__partner">{genderLine}</p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
DanilWeda, line 18: You can use isMarried inside tag, to avoid duplicating

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+};
+
+export const Person = ({ person }) => {
+  const { name, age, partnerName, isMarried, sex } = person;
+  const genderLine = `${partnerName} is my ${genderTitle[sex]}`;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
```
DanilWeda, line 13:
```suggestion
      {!!age && <p className="Person__age">{`I am ${age}`}</p>}
```
Highly important if the age is 0.
In this case, we render 0, because it will be a fallback from && operator.

DanilWeda (changes requested):
Good job. 
Check small comments. 
Thanks)

alexander-ignatow (approved):
👍

PR #2283, opened 2024-04-14
solution

ElinaNechaieva:
https://ElinaNechaieva.github.io/react_person/


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+
+export const Person = ({ person}) => {
+  
+  const { name, age, sex, isMarried, partnerName } = person;
+  return (
```
BudnikOleksii, line 5:
```suggestion
export const Person = ({ person}) => {
  const { name, age, sex, isMarried, partnerName } = person;
  
  return (
```
Add a blank line between only between statements

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+  <section className = "Person">
+    <h2 className="Person__name">My name is {name}</h2>
+    {age !== undefined && (
+    <p className="Person__age">I am {age}</p>
+    )}
```
BudnikOleksii, line 10:
```suggestion
    {age && (
      <p className="Person__age">I am {age}</p>
    )}
```
You don't need to compare with an undefined, https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+
+    {isMarried === true ? (
+       <p className="Person__partner">{partnerName} is my {sex === 'm' ? ('wife') : ('husband')}</p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
```
BudnikOleksii, line 16: - You don't need to compare with `true` as it's already boolean.
- You have `<p className="Person__partner">` in both cases, only content is different

BudnikOleksii (changes requested):
Good job 👍 
Left a few suggestions, check them

BudnikOleksii (approved):
👍 
Don't forget to deploy so reviewer can test your solution
image

PR #2287, opened 2024-04-15
add task

yanamomot:
[DEMO LINK](https://yanamomot.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,16 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerGender = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerGender = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
```
alexander-ignatow, line 8: Bug for the people with `age === 0`

alexander-ignatow (changes requested):
Very good!

Please fix one minor thing

alexander-ignatow (changes requested):
Sorry, probably I was not clear

I mean, that `0` is the valid age for the person (before the first birthday), so your component should be able to display `<p className="Person__age">I am 0</p>`, but you don't handle this age.

mrkovaliv (approved):

PR #2304, opened 2024-04-16
Develop

Olek-91:
[DEMO LINK](https://Olek-91.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  let notMarriage = `I am not married`;
+
+  if (isMarried) {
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  let notMarriage = `I am not married`;
+
+  if (isMarried) {
+    notMarriage = `${partnerName} is my ${sex === 'm' ? 'wife' : 'husband'}`;
+  }
```
DanilWeda, line 8: You can move it to a function that returns your needed text. using let in global scope isn't a good idea.
Or you can use ternary inside JSX ( worse variant)

DanilWeda (changes requested):
Good job. 
Check some comments.
Thanks!

PR #2309, opened 2024-04-17
Solution

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+  let partnerInfo = '';
```
yevhenii-pyl, line 3: Good variable naming, it actually makes sense here.

yevhenii-pyl (approved):
Good job, nice and simple 👍

PR #2310, opened 2024-04-17
Solution

lmmaarriiaa:
[DEMO LINK](https://lmmaarriiaa.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let marriageStatus = 'I am not married';
+
+  if (person.isMarried) {
+    const partner = person.sex === 'f' ? 'husband' : 'wife';
+
+    marriageStatus = `${person.partnerName} is my ${partner}`;
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+
+      {person.age !== undefined && (
+        <p className="Person__age">I am {person.age}</p>
+      )}
+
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let marriageStatus = 'I am not married';
+
+  if (person.isMarried) {
+    const partner = person.sex === 'f' ? 'husband' : 'wife';
+
+    marriageStatus = `${person.partnerName} is my ${partner}`;
+  }
```
DanilWeda, line 8: You can write a function that returns your correct text. Or you can write it inside JSX ( worse readability). Using let in global components scope worse for code support.

DanilWeda (changes requested):
Great. 
Check some small comments. 
Thanks!

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+
+      {person.age !== undefined && (
+        <p className="Person__age">I am {person.age}</p>
+      )}
```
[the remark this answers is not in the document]
lmmaarriiaa: Hello! Thank you!
I did it in case we have age: 0 and want to use it
But I will change it, thanks

src/components/Person/Person.jsx
lmmaarriiaa: Got it!
I`m not sure how to write it properly here
Could you please review my new solution?
Thanks!

mrkovaliv (approved):
Great job 👍 👍 👍

PR #2331, opened 2024-04-23
solution

YuliiaCat:
https://yuliiacat.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+
+    <p className="Person__partner">
+      {person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'f' ? 'husband' : 'wife'}`
+        : 'I am not married'}
```
etojeDenys, line 10: nested ternary operators are a bad practice, so better to move some condition to the var

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2344, opened 2024-05-01
add solution

Ellworts:


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <>
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+      {person.isMarried ? (
+        <p className="Person__partner">
+          {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
```
nazarmatsevych, line 8: Nested ternary looks complicated. Let's create a separate variable for person partner and use it here:

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married.</p>
```
nazarmatsevych, line 11: You can also have just one `<p`> tag and change only it's content

nazarmatsevych (changes requested):
Good work, please fix the comments below

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+        {person.age && <p className="Person__age">I am {person.age}</p>}
+        <p className="Person__partner">
+          {person.isMarried ? (
+            <>
+              {person.partnerName} is my {partner}
+            </>
+          ) : (
+            <>I am not married.</>
+          )}
+        </p>
```
etojeDenys, line 17:
```suggestion
        <p className="Person__partner">
          {person.isMarried ? (    
               `${person.partnerName} is my ${partner}`
            </>
          ) : (
            "I am not married."
          )}
        </p>
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = person.sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <>
```
etojeDenys, line 5: looks redundant here 

https://react.dev/reference/react/Fragment#usage

etojeDenys (changes requested):

src/components/Person/Person.jsx
Ellworts: I wrote
because the last time @nazarmatsevych checked my assignment I was asked to do this. Not sure how to solve it without "return"

lerastarynets (approved):

PR #2345, opened 2024-05-01
solution01

DanielYurchenco:
[DEMO LINK](https://DanielYurchenco.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,19 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age !== undefined && (
+      <>
+        <p className="Person__age">{`I am ${person.age}`}</p>
+      </>
+    )}
+    {person.isMarried === false ? (
+      <p className="Person__partner">I am not married</p>
+    ) : (
+      (person.sex === 'm' && (
+        <p className="Person__partner">{`${person.partnerName} is my wife`}</p>
+      )) || (
+        <p className="Person__partner">{`${person.partnerName} is my husband`}</p>
+      )
+    )}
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+        <p className="Person__age">{`I am ${person.age}`}</p>
+      </>
+    )}
+    {person.isMarried === false ? (
+      <p className="Person__partner">I am not married</p>
+    ) : (
+      (person.sex === 'm' && (
+        <p className="Person__partner">{`${person.partnerName} is my wife`}</p>
+      )) || (
+        <p className="Person__partner">{`${person.partnerName} is my husband`}</p>
```
nazarmatsevych, line 15: Let's simplifi it a bit and create a variable for person partner :

And you can also have just one <p> tag and change only it's content

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age !== undefined && (
```
nazarmatsevych, line 4:
```suggestion
    {person.age && (
```
same but shorter

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+      <>
+        <p className="Person__age">{`I am ${person.age}`}</p>
+      </>
+    )}
+    {person.isMarried === false ? (
```
nazarmatsevych, line 9:
```suggestion
    {person.isMarried === false ? (
```

You can use these shorthand

If it is false then use `!person.isMarried`

It it is true then use `person.isMarried`

nazarmatsevych (changes requested):
Good start, please check the comments below

PR #2369, opened 2024-05-20
solutionV1

kgGugala:
https://kgGugala.github.io/react_person


src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = person.sex === 'm' ? 'wife' : 'husband';
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
```
yevhenii-pyl, line 1: To make code cleaner, I'd extract all properties from the person prop at this level, to avoid typing `person.` each time.

yevhenii-pyl (approved):
Good job, keep pushing! 

For future React tasks, don't forget to extract properties from props at the highest level ✌️

PR #2373, opened 2024-05-21
Initial Commit

n3di:
[DEMO LINK](https://n3di.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = person.sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <div className="App">
+      <section className="Person">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partner = person.sex === 'm' ? 'wife' : 'husband';
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
```
yevhenii-pyl, line 1: A cleaner approach would be to extract all properties from the person here, to avoid typing `person.` each time.

yevhenii-pyl (approved):
Nice and clean! 👍 
(but don't forget to extract properties from props at the highest level)

PR #2387, opened 2024-05-25
add task solution

Julia-na:
https://Julia-na.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+
+    <p className="Person__partner">
+      {person.sex === 'f' &&
+        person.isMarried === true &&
+        `${person.partnerName} is my husband`}
+      {person.sex === 'm' &&
+        person.isMarried === true &&
+        `${person.partnerName} is my wife`}
+      {person.isMarried === false && `I am not married`}
+    </p>
+  </section>
+);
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+
+    <p className="Person__partner">
+      {person.sex === 'f' &&
+        person.isMarried === true &&
+        `${person.partnerName} is my husband`}
+      {person.sex === 'm' &&
+        person.isMarried === true &&
+        `${person.partnerName} is my wife`}
+      {person.isMarried === false && `I am not married`}
```
VitaliyBondarenko1982, line 14: try to simplify this solution.
- do not compare with boolean
- make two ternary expressions in component: first will be deside wife or husband should be rendered (let's name it partnerType) and second will be deside about final text: it will be `${person.partnerName} is my ${parnerType}` or `I am not married`

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactoring

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const partnerType = person.sex === 'f' ? 'husband' : 'wife';
+  const partnerStatus =
+    person.isMarried === true
```
vadiimvooo, line 4:
```suggestion
    person.isMarried
```

vadiimvooo (changes requested):
Almost done. Just last fix.

src/components/Person/Person.jsx
Julia-na: yes sure!

vadiimvooo (approved):
Well done

PR #2396, opened 2024-05-29
solution

yaros-dev:
[DEMO LINK](https://yaros-dev.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {!!age && <p className="Person__age">I am {age}</p>}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${sex === 'f' ? 'husband' : 'wife'}`
+          : 'I am not married'}
```
etojeDenys, line 11: nested ternary operators are a bad practice



```
```

etojeDenys (changes requested):

src/components/Person/Person.jsx
yaros-dev: Thank you! I already had experience when data was not correctly calculated during the rendering of a component due to nested ternary operators. Putting it in a variable helped.

etojeDenys (approved):
🔥

PR #2410, opened 2024-06-07
person component

MariCodec:
[DEMO LINK](https://MariCodec.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.sex === 'm'
+          ? `${person.partnerName} is my wife`
+          : `${person.partnerName} is my husband`}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married </p>
+    )}
+  </section>
+);
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@ and use it 3 times inside the `App` instead of static markup.
 - keep the same classNames in `Person.jsx` as in `App.jsx` (`Person`, `Person__name`, `Person__age`, `Person__partner`).
 
 ## Instructions
+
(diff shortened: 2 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
roman-mirzoian, line 1: Let's add such a constant to avoid magic values.

```suggestion
const MALE = 'm';
const FEMALE = 'f';

export const Person = ({ person }) => (
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  <section className="Person">
```
roman-mirzoian, line 2: Also possible to transfer the expression about the definition of the partner to a separate constant.
Example:

```suggestion
export const Person = ({ person }) => (
  const partnerRole = person.sex === 'm' ? 'wife' : 'husband';
  
  <section className="Person">
```

roman-mirzoian (changes requested):
Good job here are the comments to improve your code.

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+const MALE = 'm';
+const partner = sex => (sex === MALE ? 'wife' : 'husband');
+
+export const Person = ({ person }) => (
```
dinaminahi, line 4: Let's destructure properties from person to improve code readability.

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
...
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {`${person.partnerName} is my ${partner(person.sex)}`}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married </p>
```
dinaminahi, line 13: We are repeating <p className="Person__partner"> 2 times . Let's do the check inside a p tag so that will have it only once and I would move it to a variable for better readability

```
const someVar = check here
<p>{someVar}</p>
```

dinaminahi (changes requested):
Good job 👍 Let's do some fixes to improve a bit

dinaminahi (approved):
Well done ⭐

PR #2412, opened 2024-06-07
Solution

OSpyrydonov:
[DEMO LINK](https://OSpyrydonov.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.sex === 'm'
+          ? `${person.partnerName} is my wife`
+          : `${person.partnerName} is my husband`}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married </p>
+    )}
+  </section>
+);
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
roman-mirzoian, line 1: Let's add such a constant to avoid magic values.

```
const MALE = 'm';
const FEMALE = 'f';
```

roman-mirzoian, line 1: Also possible to transfer the expression about the definition of the partner to a separate constant.
Example:

```suggestion
export const Person = ({ person }) => (
  const partnerRole = sex === MALE ? 'wife' : 'husband';
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
```
roman-mirzoian, line 3: Here you can use destructuring to improve code readability.

Usage example:

```suggestion
  <section className="Person">
    const { name, age, sex, isMarried, partnerName } = person;
    
    <h2 className="Person__name">{`My name is ${name}`}</h2>
```

roman-mirzoian (changes requested):
Good job, here are the comments to improve your code.

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+      {isMarried ? (
+        <p className="Person__partner">
+          {`${partnerName} is my ${partnerRole}`}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
dinaminahi, line 25: We are repeating <p className="Person__partner"> 2 times . Let's do the check inside a p tag so that will have it only once and I would move it to a variable for better readability

dinaminahi (changes requested):
Good job 👍 But let's do a small fix

dinaminahi (approved):
Well done ⭐

PR #2418, opened 2024-06-09
my solution

RuslanHnatyshyn:
[DEMO LINK](https://RuslanHnatyshyn.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <React.Fragment>
+      <section className="Person">
+        <h2 className="Person__name">My name is {name}</h2>
+        {age === undefined ? null : <p className="Person__age">I am {age}</p>}
+        {isMarried ? (
+          <p className="Person__partner">
+            {partnerName} is my {sex === 'm' ? 'wife' : 'husband'}
+          </p>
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+import React from 'react';
+
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <React.Fragment>
+      <section className="Person">
+        <h2 className="Person__name">My name is {name}</h2>
+        {age === undefined ? null : <p className="Person__age">I am {age}</p>}
```
dinaminahi, line 10: You can use logical operator && here 
```suggestion
        {!!age && <p className="Person__age">I am {age}</p>}
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+        {isMarried ? (
+          <p className="Person__partner">
+            {partnerName} is my {sex === 'm' ? 'wife' : 'husband'}
+          </p>
+        ) : (
+          <p className="Person__partner">I am not married</p>
+        )}
```
dinaminahi, line 17: Let's do the check inside a p tag so that will have it only once and I would move it to a variable for better readability
```
const someVar = check here
<p>{someVar}</p>
```

dinaminahi (changes requested):
Great job 👍 Let's improve it a bit

roman-mirzoian (approved):
Well done!

RuslanHnatyshyn:
Thank you for your comments.

PR #2420, opened 2024-06-09
add person solution

FaiHamid:
[DEMO LINK](https://FaiHamid.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,19 @@
 // export const Person = ({ person }) => ();
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age !== undefined && (
+      <p className="Person__age">I am {person.age}</p>
+    )}
+
+    {person.isMarried === true ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner"> I am not married</p>
+    )}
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
 // export const Person = ({ person }) => ();
+
+export const Person = ({ person }) => (
```
dinaminahi, line 3: Let's destructure properties from person

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age !== undefined && (
```
dinaminahi, line 7: Wat if it is `null`? let's just do `!!age &&`

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+      <p className="Person__age">I am {person.age}</p>
+    )}
+
+    {person.isMarried === true ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner"> I am not married</p>
+    )}
```
dinaminahi, line 17: Let's do the check inside a p tag so that will have it only once and I would move it to a variable for better readability
```
const someVar = check here
<p>{someVar}</p>
```

dinaminahi (changes requested):
Good job 👍 Let's improve it a bit

roman-mirzoian (approved):
Well done!

PR #2421, opened 2024-06-09
add solution person

OleksiiKnihin:
[DEMO LINK](https://OleksiiKnihin.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age ? <p className="Person__age">I am {age}</p> : null}
+      {isMarried ? (
+        <p className="Person__partner">
+          {partnerName} is my {partner}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
+    </section>
+  );
+};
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age ? <p className="Person__age">I am {age}</p> : null}
+      {isMarried ? (
+        <p className="Person__partner">
+          {partnerName} is my {partner}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
```
dinaminahi, line 14: Let's do the check inside a p tag so that will have it only once, because now we are repeating the tag and its classname

dinaminahi (changes requested):
Good job 👍 Let's improve it a bit

src/components/Person/Person.jsx
OleksiiKnihin: OK. I don`t mind)

OleksiiKnihin:
Fix code after mentor`s remark

src/components/Person/Person.jsx
```diff
@@ -1 +1,14 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partner = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age ? <p className="Person__age">I am {age}</p> : null}
```
roman-mirzoian, line 8: Here you can improve code by using logical AND `&&` instead of ternary operator with `null`. 
This will make the code a little more compact.

```suggestion
      {age && <p className="Person__age">I am {age}</p>}
```

roman-mirzoian (approved):
Well done and please pay attention to the comment.

PR #2425, opened 2024-06-10
solution

Stasandrushchak:
[DEMO LINK](https://Stasandrushchak.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerLabel = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      {isMarried ? (
+        <p className="Person__partner">
+          {partnerName} is my {partnerLabel}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
dinaminahi, line 15: We are repeating `<p className="Person__partner">` 2 times . Let's do the check inside a p tag so that will have it only once and I would move it to a variable for better readability

dinaminahi (changes requested):
Good job 👍 But let's do a small fix

dinaminahi (approved):
Well done 💪

PR #2426, opened 2024-06-10
solution

priemssonia:


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    {person.isMarried &&
+      (person.sex === 'm' ? (
+        <p className="Person__partner">{person.partnerName} is my wife</p>
+      ) : (
+        <p className="Person__partner">{person.partnerName} is my husband</p>
+      ))}
+    {!person.isMarried && <p className="Person__partner">I am not married</p>}
```
dinaminahi, line 11: We are repeating <p className="Person__partner"> 3 times . Let's do the check inside a p tag so that will have it only once and I would move it to a variable for better readability
Also let's add some constants to avoid magic values
const MALE = 'm';
const FEMALE = 'f';

src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
```
dinaminahi, line 1: Let's destructure properties from person to improve code readability.

dinaminahi (changes requested):
Good job 👍 Let's do some fixes to improve a bit

dinaminahi (approved):
Well done ⭐

PR #2440, opened 2024-06-20
solution but with bugs, need help

akryto:
- if a person is not marriend - print `I am not married` message;
- keep the same classNames in `Person.jsx` as in `App.jsx` (`Person`, `Person__name`, `Person__age`, `Person__partner`).
[DEMO LINK](https://akryto.github.io/react_person/)


roman-mirzoian (changes requested):
Good work here are the comments to improve the code.

PR #2443, opened 2024-06-29
add task solution

ashatskyy:
[DEMO LINK](https://ashatskyy.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,11 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {'age' in person ? <p className="Person__age">I am {person.age}</p> : null}
+    <p className="Person__partner">
+      {person.isMarried === true
```
etojeDenys, line 6:
```suggestion
      {person.isMarried
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,11 @@
...
+        ? `${person.partnerName} is my ${person.sex === 'm' ? `wife` : `husband`}`
+        : `I am not married`}
```
etojeDenys, line 8: nested ternary operators are a bad practice


```
```

etojeDenys (changes requested):

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const personPartnerType = person.sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
```
volodymyr-soltys97, line 6: Use destructuring for `person`

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+      {'age' in person ? (
+        <p className="Person__age">I am {person.age}</p>
+      ) : null}
```
volodymyr-soltys97, line 9: Fix it after destructuring
```suggestion
      {age && (
        <p className="Person__age">I am {age}</p>
      )}
```

volodymyr-soltys97 (changes requested):
GJ!
Let's make your code better

src/components/Person/Person.jsx
ashatskyy: Good evening! In this task instructions we see condition: ''pass the whole person as a prop <Person person={misha} /> (not individual fields)'' Is your comment matches this condition?

PR #2452, opened 2024-07-07
solution

VSkits:
[DEMO LINK](https://VSkits.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age !== undefined && <p className="Person__age">I am {age}</p>}
+      {isMarried ? (
+        <p className="Person__partner">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age !== undefined && <p className="Person__age">I am {age}</p>}
```
TarasHoliuk, line 7:
```suggestion
      {age && <p className="Person__age">I am {age}</p>}
```
Or you can explicitly transform it into a boolean
```suggestion
      {!!age && <p className="Person__age">I am {age}</p>}
```

But still, it will look more laconic

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+      {isMarried ? (
+        <p className="Person__partner">
```
TarasHoliuk, line 9: tag and className is same in both cases
So you can put ternary inside this tag to define different content only

TarasHoliuk (changes requested):
GJ 👍 
2 small suggestions

anastasiiavorobiova (approved):
Great work!

PR #2456, opened 2024-07-08
solution

IrbisKronos:
[DEMO LINK](https://IrbisKronos.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,13 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,13 @@
...
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">I am {person.age}</p>}
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'f' ? 'husband' : 'wife'}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
+    )}
```
etojeDenys, line 11: nested ternary operators are a bad practice, better to move that logic to some var

etojeDenys:
```
```

etojeDenys (changes requested):

PR #2460, opened 2024-07-14
add solution

RainerWerner44:
[DEMO LINK](https://RainerWerner44.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({
+  person: { name, age = null, sex, isMarried, partnerName },
+}) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {name}</h2>
+    {age !== null && <p className="Person__age">I am {age}</p>}
+    <p className="Person__partner">
+      {isMarried === true && sex === 'm' ? <p>{partnerName} is my wife</p> : ''}
+      {isMarried === true && sex === 'f' ? (
+        <p>{partnerName} is my husband</p>
+      ) : (
+        ''
+      )}
+      {isMarried === false ? <p>I am not married</p> : ''}
+    </p>
+  </section>
+);
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Person } from './components/Person/Person';
 
 export const misha = {
   name: 'Misha',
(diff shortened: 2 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({
+  person: { name, age = null, sex, isMarried, partnerName },
+}) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {name}</h2>
+    {age !== null && <p className="Person__age">I am {age}</p>}
+    <p className="Person__partner">
+      {isMarried === true && sex === 'm' ? <p>{partnerName} is my wife</p> : ''}
```
ArturErmolenko, line 8: isMarried is already boolean why you need this check?
Please use correct formatting, can we use variables here?

src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
...
+      {isMarried === true && sex === 'f' ? (
+        <p>{partnerName} is my husband</p>
+      ) : (
+        ''
+      )}
```
ArturErmolenko, line 13:
```suggestion
      {isMarried && sex === 'f' 
        ? (
          <p>{partnerName} is my husband</p>
        ) 
        : (
          ''
        )}
```

Formatting, variables, indents

src/components/Person/Person.jsx
```diff
@@ -1 +1,26 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age = undefined, sex, isMarried, partnerName } = person;
+
+  const ageText =
+    age !== undefined ? <p className="Person__age">I am {age}</p> : null;
```
BudnikOleksii, line 5: don't need to set default value as undefined, and strictly compare with it. Just check if a value exists and actually you don't need to set null. In react you can render conditionally directly in jsx:
 ```jsx
```

File: src/components/Person/Person.jsx [hunk 4]
```diff
@@ -1 +1,26 @@
...
+
+  if (isMarried) {
+    if (sex === 'm') {
+      partnerText = `${partnerName} is my wife`;
+    } else if (sex === 'f') {
+      partnerText = `${partnerName} is my husband`;
+    }
+  } else {
+    partnerText = 'I am not married';
+  }
```
BudnikOleksii, line 17: - Avoid nested `if/else` statements, and `else if` at all.
- 'm' and 'f' are magic values, create constants for them

BudnikOleksii (changes requested):
Good work 👍 
But could be even better, check the comments

DanilWeda (approved):
Good job. 
Approved! 
Many thanks)

PR #2468, opened 2024-07-19
sol

DenPrislipskyi:
[DEMO LINK](https://DenPrislipskyi.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  let partnerMessage;
+
+  if (person.isMarried) {
+    if (person.sex === 'm') {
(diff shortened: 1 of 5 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+
+  if (person.isMarried) {
+    if (person.sex === 'm') {
+      partnerMessage = `${person.partnerName} is my wife`;
+    } else {
+      partnerMessage = `${person.partnerName} is my husband`;
+    }
+  } else {
+    partnerMessage = 'I am not married';
+  }
```
BudnikOleksii, line 12: Let's try to get rid of so many if/else statements

BudnikOleksii (changes requested):
Good work 👍 
Let's try to refactor and get rid of if/else statements

PR #2473, opened 2024-07-23
add task solution

YaroslavKolesnyk:
[DEMO LINK](https://YaroslavKolesnyk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  function isMaried(people) {
+    if (!people.isMarried) {
+      return `I am not married`;
+    }
+
+    const partner = people.sex === 'f' ? 'husband' : 'wife';
+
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  function isMaried(people) {
```
BudnikOleksii, line 2: - Move this function outside the component
- Function name should start with a verb. And as it returns a string, it shouldn't be something with `is` as `is` or `has` are used for a boolean variables

BudnikOleksii (changes requested):
Please, deploy your solution so a reviewer can test the UI
image

src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
-// export const Person = ({ person }) => ();
+function getMaritalStatus(people) {
```
BudnikOleksii, line 1: Looks like it's a single person

BudnikOleksii (approved):
👍

PR #2476, opened 2024-07-23
add solution

iyehorova:
[DEMO LINK](https://iyehorova.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,17 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age = null, sex, isMarried, partnerName = null } = person;
```
BudnikOleksii, line 2: You don't need to set default values as null here, it doesn't make much sense

BudnikOleksii (approved):
Good work 👍

PR #2479, opened 2024-07-24
Develop

RimmaBuzuluk:
https://RimmaBuzuluk.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age ? <p className="Person__age">I am {person.age}</p> : ''}
+      {person.isMarried ? (
+        <p className="Person__partner">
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <div className="App">
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age ? <p className="Person__age">I am {person.age}</p> : ''}
```
BudnikOleksii, line 5: You can use `&&` operator

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+      {person.isMarried ? (
+        <p className="Person__partner">
```
BudnikOleksii, line 7: `<p className="Person__partner">` in both cases, only the content is different

BudnikOleksii (changes requested):
Good work 👍 
Just a few small changes are required

src/components/Person/Person.jsx
RimmaBuzuluk: I can not delete "Person__partner", because it is broke test

BudnikOleksii (approved):
👍

PR #2483, opened 2024-07-24
React Person Task Solution

piyopiyo0:
[DEMO LINK](https://piyopiyo0.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
...
+    <h2 className="Person__name">My name is {name}</h2>
+
+    {age && <p className="Person__age">I am {age}</p>}
+    {isMarried && sex === 'm' && (
+      <p className="Person__partner">{partnerName} is my wife</p>
+    )}
+    {isMarried && sex === 'f' && (
+      <p className="Person__partner">{partnerName} is my husband</p>
+    )}
+    {!isMarried && <p className="Person__partner">I am not married</p>}
```
DanilWeda, line 14: Try using one condition 
IsMarried ? condition which render wife or husband : default text ( I am not.. )

DanilWeda (changes requested):
Good job. 
Check a comment for improving your solution. 
Many thanks)

src/components/Person/Person.jsx
piyopiyo0: Fixed, thanks a lot)

PR #2493, opened 2024-07-25
added solution

NIKaragu:
[DEMO LINK](NIKaragu.github.io/react_person/)

### DIFF
--- a/src/components/Partner/Partner.jsx
+++ b/src/components/Partner/Partner.jsx
@@ -1,0 +1,15 @@
+export const Partner = ({ person }) => {
+  const { partnerName, sex, isMarried } = person;
+
+  return (
+    <>
+      {isMarried ? (
+        <p className="Person__partner">
+          {partnerName} is my {sex === 'f' ? 'husband' : 'wife'}
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
+    </>
+  );
+};
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,16 @@
-// export const Person = ({ person }) => ();
+import { Partner } from '../Partner/Partner';
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name, age } = person;
+
+  return (
+    <>
+      <section className="Person">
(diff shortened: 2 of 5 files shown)


src/components/Person/Person.scss
```diff
@@ -0,0 +1,16 @@
+.Person {
```
DanilWeda, line 1: For class name better use lowercase

src/components/Person/Person.jsx
```diff
@@ -1 +1,16 @@
-// export const Person = ({ person }) => ();
+import { Partner } from '../Partner/Partner';
+import './Person.scss';
+
+export const Person = ({ person }) => {
+  const { name, age } = person;
+
+  return (
+    <>
```
DanilWeda, line 8: You don`t need jsx fragment in this case ( you have section as parent tag element)

src/components/Partner/Partner.jsx
```diff
@@ -0,0 +1,15 @@
+export const Partner = ({ person }) => {
+  const { partnerName, sex, isMarried } = person;
+
+  return (
+    <>
+      {isMarried ? (
+        <p className="Person__partner">
```
DanilWeda, line 7: Also change class name in jsx to lowercase

src/components/Partner/Partner.jsx
DanilWeda, line 1: Great! 
Many thanks for moving logic with the person in a separate component.

DanilWeda (changes requested):
Good job. 
Check some small minor comments. 
Your solution is awesome! 
Many thanks)

src/components/Person/Person.scss
NIKaragu: I can't, 'cause tests try to catch these classes
image

src/components/Partner/Partner.jsx
NIKaragu: answer is here https://github.com/mate-academy/react_person/pull/2493#discussion_r1692218728

src/components/Person/Person.jsx
NIKaragu: done

BudnikOleksii (approved):
👍

PR #2494, opened 2024-07-25
Solution

totum-revolutum:
[DEMO LINK](https://totum-revolutum.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const spouse = sex === 'f' ? 'husband' : 'wife';
+  let spouseStatus = <p className="Person__partner">I am not married</p>;
+
+  if (isMarried) {
+    spouseStatus = (
+      <p className="Person__partner">
+        {partnerName} is my {spouse}
+      </p>
+    );
+  }
+
+  return (
+    <section className="Person">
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+  const spouse = sex === 'f' ? 'husband' : 'wife';
+  let spouseStatus = <p className="Person__partner">I am not married</p>;
+
+  if (isMarried) {
+    spouseStatus = (
+      <p className="Person__partner">
+        {partnerName} is my {spouse}
+      </p>
+    );
+  }
```
BudnikOleksii, line 12: - It's better to use const and you can do a check for `isMarried` directly in return statement instead of assign to the variable
- `<p className="Person__partner">` in both cases, only the content is different

BudnikOleksii (changes requested):
Good work 👍 
Just a few minor fixes are required

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const spouse = sex === 'f' ? 'husband' : 'wife';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      {isMarried ? (
+        <p className="Person__partner">
```
BudnikOleksii, line 10: `<p className="Person__partner">` in both cases, only the content is different, so don't repeat it

BudnikOleksii (approved):
👍

PR #2495, opened 2024-07-25
Solution task

SashaPas05:
[DEMO LINK](https://SashaPas05.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,19 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+
+      {age && <p className="Person__age">I am {age}</p>}
+
+      {isMarried ? (
+        <p className="Person__partner">
+          {partnerName} is my {sex === 'm' ? `wife` : `husband`}
+        </p>
+      ) : (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+
+      {age && <p className="Person__age">I am {age}</p>}
+
+      {isMarried ? (
+        <p className="Person__partner">
```
BudnikOleksii, line 11: `<p className="Person__partner">` in both cases, only the content is different

BudnikOleksii (changes requested):
Please, deploy your solution so a reviewer can test it

src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+          {partnerName} is my {sex === 'm' ? `wife` : `husband`}
```
ArturErmolenko, line 12:
```suggestion
          {partnerName} is my {sex === 'm' ? `wife` : `husband`}
```
let's create variable for this expression sex === 'm' ? `wife` : `husband`
`Also  <p className="Person__partner"> in both cases, only the content is different`
this comment didn't fixed, use only one `<p>` and change only content you don't need to repeat `<p>` two times

ArturErmolenko (changes requested):
Almost done 🔥 
Let's improve a bit

PR #2498, opened 2024-07-26
Develop

YarikKedrych:


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+export const Person = ({ person }) => {
+  const marriedLable = person.sex === 'f' ? 'husband' : 'wife';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {!!person.age && <p className="Person__age">I am {person.age}</p>}
+
+      {person.isMarried ? (
+        <p className="Person__partner">
```
BudnikOleksii, line 10: `<p className="Person__partner">` in both cases, only the content is different, so don't repeat it

BudnikOleksii (changes requested):
Good work 👍 
Just one minor fix is required

src/components/Person/Person.jsx
YarikKedrych: class 'person_partner' repeated  in case ' I am  not married ' ?

src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const marriedLable = person.sex === 'f' ? 'husband' : 'wife';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
```
ArturErmolenko, line 6: It will be great to destructing name age ... from person

ArturErmolenko (approved):
Nice 👍

PR #2503, opened 2024-07-28
person

NavI-Rap0:
https://NavI-Rap0.github.io/react_person/

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,21 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+
+  let partnersText = 'I am not married';
+
+  if (isMarried === true && sex === 'm') {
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,21 @@
...
+
+  let partnersText = 'I am not married';
+
+  if (isMarried === true && sex === 'm') {
+    partnersText = `${partnerName} is my wife`;
+  } else if (isMarried === true && sex === 'f') {
+    partnersText = `${partnerName} is my husband`;
+  } else {
+    partnersText = 'I am not married';
+  }
```
BudnikOleksii, line 12: - Try to simplify it and get rid of `if`/`else if`/`else` statements
- You don't need to compare with boolean variables as `isMarried` is already boolean

BudnikOleksii (changes requested):
Good work 👍 
Let's fix a few moments

BudnikOleksii (approved):
👍

PR #2550, opened 2024-09-07
feat: person component

Ann-Maksay:
[DEMO LINK](https://Ann-Maksay.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,19 @@
 // export const Person = ({ person }) => ();
+
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+
+    {person.age !== undefined && (
+      <p className="Person__age">I am {person.age}</p>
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,19 @@
...
+    {person.age !== undefined && (
+      <p className="Person__age">I am {person.age}</p>
+    )}
+
+    {person.isMarried ? (
+      <p className="Person__partner">
+        {person.partnerName} is my {person.sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
+      <p className="Person__partner">I am not married</p>
```
BudnikOleksii, line 16: `<p className="Person__partner">` in both cases, only content is different

BudnikOleksii (changes requested):
Good work 👍 
Just one moment should be fixed, and something wrong with demo, I don't see any UI

BudnikOleksii (approved):
Looks good 👍

PR #2555, opened 2024-09-10
solved task

PikalovAndrey:
[DEMO LINK](https://PikalovAndrey.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,22 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+    let partnerText;
+  
+    if (person.isMarried) {
+      partnerText = person.sex === 'f'
+        ? `${person.partnerName} is my husband`
+        : `${person.partnerName} is my wife`;
+    } else {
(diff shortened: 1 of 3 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,22 @@
...
+export const Person = ({ person }) => {
+    let partnerText;
+  
+    if (person.isMarried) {
+      partnerText = person.sex === 'f'
+        ? `${person.partnerName} is my husband`
+        : `${person.partnerName} is my wife`;
+    } else {
+      partnerText = 'I am not married';
+    }
```
BudnikOleksii, line 10: Try to simplify this logic to avoid repeating and `else` statements

BudnikOleksii (changes requested):
Good work 👍 
Let's try to simplify a bit

src/components/Person/Person.jsx
PikalovAndrey: Firstly i tried to do this, but linter doesn't like nested ternary.

polosanya (approved):

PR #2587, opened 2024-09-18
Add task solution

D-Ptashevskyi:
https://D-Ptashevskyi.github.io/react_person/

### DIFF
--- a/src/Person.jsx
+++ b/src/Person.jsx
@@ -1,19 +1,16 @@
 export const Person = ({ person }) => {
   const { name, age, sex, isMarried, partnerName } = person;
 
+  const partnerStatus =
+    sex === 'm' ? `${partnerName} is my wife` : `${partnerName} is my husband`;
+
   return (
     <section className="Person">
       <h2 className="Person__name">My name is {name}</h2>
       {age && <p className="Person__age">I am {age}</p>}
-      {isMarried ? (
-        <p className="Person__partner">
-          {sex === 'm'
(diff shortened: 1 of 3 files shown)


src/Person.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      {isMarried ? (
+        <p className="Person__partner">
+          {sex === 'm'
+            ? `${partnerName} is my wife`
+            : `${partnerName} is my husband`}
```
igorkruz: create a variable for this logic

src/Person.jsx
```diff
@@ -0,0 +1,19 @@
...
+        </p>
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
igorkruz: you can simplify this and render just one `p`  and show the data according to the condition

```suggestion
         <p className="Person__partner">
            {isMarried
              ? -----------------
              : 'I am not married'}
         </p>
```

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

Add the `DEMO LINK` with your solution

igorkruz (changes requested):
Looks good 👍 
 you should add the `DEMO LINK` with your solution

igorkruz (approved):
Well done 🔥

PR #2597, opened 2024-09-24
add task solution

BohdanDymydiuk:
[DEMO LINK](https://BohdanDymydiuk.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,11 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    <p className="Person__partner">{`${
+      person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+        : 'I am not married'
+    }`}</p>
+  </section>
+);
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,11 +1,15 @@
-export const Person = ({ person }) => (
-  <section className="Person">
-    <h2 className="Person__name">My name is {person.name}</h2>
-    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
-    <p className="Person__partner">{`${
-      person.isMarried
-        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,11 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
```
volodymyr-soltys97: Use destructuring for `person`

src/components/Person/Person.jsx
```diff
@@ -1 +1,11 @@
...
+    {person.age && <p className="Person__age">{`I am ${person.age}`}</p>}
+    <p className="Person__partner">{`${
+      person.isMarried
+        ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+        : 'I am not married'
+    }`}</p>
```
volodymyr-soltys97: Move this logic to the helper variables and use it here

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/components/Person/Person.jsx
BohdanDymydiuk: I don't have to do this beacuse of that:
image

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const isPersonParried = `${
```
volodymyr-soltys97:
```suggestion
  const isPersonParried = 
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+    person.isMarried
+      ? `${person.partnerName} is my ${person.sex === 'm' ? 'wife' : 'husband'}`
+      : 'I am not married'
+  }`;
```
volodymyr-soltys97:
```suggestion
      : 'I am not married';
```

src/components/Person/Person.jsx
volodymyr-soltys97: Use destructuring for `person`
```suggestion
export const Person = ({ person }) => {
  const { age, .......... } = person;

  const isPersonParried = `${
```

volodymyr-soltys97 (changes requested):
Almost done!
This comment still not fixed
You don't need to change props, you need only use destructuring for this props object 
image

lerastarynets (approved):
lgtm 🚀

PR #2600, opened 2024-09-26
Solution

anna-agerone:
[DEMO LINK](https://anna-agerone.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,15 +1,18 @@
-export const Person = ({
-  person: { name, age, sex, isMarried, partnerName },
-}) => (
-  <section className="Person">
-    <h2 className="Person__name">My name is {name}</h2>
-    {age && <p className="Person__age">I am {age}</p>}
-    {isMarried ? (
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
...
+  person: { name, age, sex, isMarried, partnerName },
+}) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {name}</h2>
+    {age && <p className="Person__age">I am {age}</p>}
+    {isMarried ? (
+      <p className="Person__partner">
+        {partnerName} is my {sex === 'm' ? 'wife' : 'husband'}
+      </p>
+    ) : (
```
etojeDenys: better to avoid nested ternary operators 

```
```

etojeDenys (approved):
lgtm

src/components/Person/Person.jsx
anna-agerone: Thank you! I updated my .jsx file.

PR #2698, opened 2024-11-19
done task

Jion353:


src/components/Person/Person.jsx
```diff
@@ -1 +1,15 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => (
+  <section className="Person">
+    <h2 className="Person__name">My name is {person.name}</h2>
+    {person.age ? (
+      <p className="Person__age"> {`I am ${person.age}`} </p>
+    ) : null}
```
BudnikOleksii, line 6:
```suggestion
    {person.age && (
      <p className="Person__age"> {`I am ${person.age}`} </p>
    )}
```
You don't need to use ternary operator if you return nothing in the second case

BudnikOleksii (changes requested):
Good work 👍 
Approved, but left a comments and don't forget to add a demo link so a reviewer can test your solution

PR #2706, opened 2024-11-20
task solution

AnderZerfall:
[DEMO LINK](https://AnderZerfall.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,1 +1,23 @@
 // export const Person = ({ person }) => ();
+
+export const Person = ({ person }) => {
+  let partnerSex = '';
+
+  if (person.sex === 'm') {
+    partnerSex = 'wife';
+  } else {
+    partnerSex = 'husband';
+  }
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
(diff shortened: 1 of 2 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
 // export const Person = ({ person }) => ();
+
+export const Person = ({ person }) => {
+  let partnerSex = '';
+
+  if (person.sex === 'm') {
+    partnerSex = 'wife';
+  } else {
+    partnerSex = 'husband';
+  }
```
BudnikOleksii, line 10: You can do it with ternary operator

src/components/Person/Person.jsx
```diff
@@ -1 +1,23 @@
...
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {person.name}</h2>
+      {person.age && <p className="Person__age">I am {person.age}</p>}
+      <p className="Person__partner">
+        {person.isMarried === true
```
BudnikOleksii, line 17: You don't need to compare with true, it's already boolean

BudnikOleksii (changes requested):
Good work 👍 
Let's clean your solution a bit

PR #2765, opened 2024-12-22
solution

djtyrf312:


src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+  <section className="Person">
+    <h2 className="Person__name">{`My name is ${person.name}`}</h2>
+    <Age age={person.age} />
+    <Partner
+      isMarried={person.isMarried}
+      sex={person.sex}
+      partnerName={person.partnerName}
+    />
+  </section>
+);
```
naziamoff: looks good actually, let's see the alternative formatting:

```suggestion
export const Person = ({ person }) => {
  const { age, sex, partnerName, isMarried } = person;

    return (
      <section className="Person">
        <h2 className="Person__name">
          {`My name is ${name}`}
        </h2>
        
        <Age age={age} />
        
        <Partner
          isMarried={isMarried}
          sex={sex}
          partnerName={partnerName}
        />
      </section>
    );
};
```

It depends on the formatting conventions of the company, but I prefer it this way.  What changed:
- added `person` object destructuring in order to shorten the syntax when passing to props
- added one line in between components of the same nesting level
- moved the text from `h2` to a separate line in order to improve readability

For now we can treat it as a matter of taste, you may choose what looks better for you

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+
```
naziamoff, line 7: we usually use separate files for each component. 
So there should be `Age.tsx` and `Partner.tsx`, which are exported from their own files. I'd suggest you do it too

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+const Age = ({ age }) => age && <p className="Person__age">{`I am ${age}`}</p>;
```
naziamoff: Here's an alternative:

```suggestion
const Age = ({ age }) => {
  if (!age) {
    return null;
  }

return (
  <p className="Person__age">
    {`I am ${age}`}
  </p>
)};
```

I prefer this way more, as it looks more readable and clear to me. But again, depends heavily on the formatting conventions of the project

src/components/Person/Person.jsx
```diff
@@ -1 +1,28 @@
...
+  if (isMarried && sex === 'm') {
+    message = `${partnerName} is my wife`;
+  }
+
+  if (isMarried && sex === 'f') {
+    message = `${partnerName} is my husband`;
+  }
+
+  return <p className="Person__partner">{`${message}`}</p>;
+};
```
naziamoff: If `isMarried` is false - we know for sure that we should return `I am not married` and we don't need to do further checks

The same is true for `isMarried && sex === 'm'`— we already know what we need to return and the `message` won't be reassigned, since the conditions exclude each other.

In the current code version, we run all the checks no matter what. I suggest to think of a way to avoid `let` here and return right away

naziamoff (approved):
⭐

PR #2786, opened 2025-01-09
solution

Blervin1:
[DEMO LINK](https://Blervin1.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
...
+  const { name, age, isMarried, sex, partnerName } = person;
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      <p className="Person__partner">
+        {isMarried
+          ? `${partnerName} is my ${sex === 'm' ? 'wife' : 'husband'}`
+          : 'I am not married'}
```
etojeDenys, line 14: good job, but it's better to avoid nested ternary operator

etojeDenys: decide before jsx what text will be rendered and just render in jsx. something like that

```
```

etojeDenys (approved):

PR #2964, opened 2025-03-14
Solution

Joao-Campos06:
[DEMO LINK](https://Joao-Campos06.github.io/react_person/)


src/components/Person/Person.jsx
```diff
@@ -1 +1,18 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const partnerRole = sex === 'm' ? 'wife' : 'husband';
+
+  return (
+    <section className="Person">
+      <h2 className="Person__name">My name is {name}</h2>
+      {age && <p className="Person__age">I am {age}</p>}
+      {isMarried ? (
```
DanielCamposDevX, line 9: On this ternary the only change you made is the text inside the p tag.
AI:   Consider creating a separate variable for the condition in the Person__partner element's conditional rendering.

The AI said that because the folowing conditions whould be simply met by using something like:
Const exampleText = conditional ? "My first text" : "My second text";
In this way you can use the variable directly here on the <p>

DanielCamposDevX (changes requested):
AI: Consider creating a separate variable for the condition in the Person__partner element's conditional rendering.

The AI said that because the folowing conditions whould be simply met by using something like:
Const exampleText = conditional ? "My first text" : "My second text";
In this way you can use the variable directly here on the

PR #3052, opened 2025-05-06
add task solution

hannachachacha:
- [DEMO LINK](https://hannachachacha.github.io/react_person/)


hannachachacha:
@mateacademy-ai-mentor but in the readme.md it says: "- omit `age` if it is not given;" and Olya's age is not given

PR #3064, opened 2025-05-12
task solution

nbryk:
[DEMO LINK](https://nbryk.github.io/react_person/)


nbryk:
According to the terms of the assignment, I cannot change the objects.

ArturErmolenko (approved):

PR #3070, opened 2025-05-13
add solution

zbvika:
[DEMO LINK](https://zbvika.github.io/react_person/)


zbvika:
@mateacademy-ai-mentor Thank you for the feedback.
According to the task, Olya is not supposed to have an age property, as this is part of a test case. We shouldn't add the age manually—instead, we should avoid displaying age information if the property is missing.

As for the second point, if I understood correctly, my logic already matches the requirement:

if the person is male (sex === 'm'), their partner is labeled "wife"

if the person is female (sex === 'f'), their partner is labeled "husband"

If I misunderstood the comment, please clarify.

PR #3075, opened 2025-05-13
Add task solution

hryhoriikorsun:
[DEMO LINK](https://hryhoriikorsun.github.io/react_person/)

### DIFF
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -15,7 +15,7 @@ export const Person = ({ person }) => {
     <section className="Person">
       <h2 className="Person__name">{`My name is ${name}`}</h2>
 
-      {age !== undefined && <p className="Person__age">{`I am ${age}`}</p>}
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
 
       {isMarried === true ? (
         sexOfPartner
--- a/src/components/Person/Person.jsx
+++ b/src/components/Person/Person.jsx
@@ -1,15 +1,12 @@
 export const Person = ({ person }) => {
   const { name, age, sex, isMarried, partnerName } = person;
-
-  let sexOfPartner = (
-    <p className="Person__partner">{`${partnerName} is my wife`}</p>
-  );
-
-  if (sex === 'f') {
-    sexOfPartner = (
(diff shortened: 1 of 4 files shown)


src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+
+  let sexOfPartner = (
+    <p className="Person__partner">{`${partnerName} is my wife`}</p>
+  );
+
+  if (sex === 'f') {
+    sexOfPartner = (
+      <p className="Person__partner">{`${partnerName} is my husband`}</p>
+    );
+  }
```
BudnikOleksii: - It's better to create a separate function to get this text to avoid additional logic in the component
- 'f' is a magic value
- You repeat logic <p className="Person__partner">{`${partnerName} is my husband`}</p>

You can simplify like this
```
const personPartnerText = isMarried ?  `${partnerName} is my ${sex === SEX.female ? 'husband' : 'wife'}` : DEFAULT_PERSON_PARTNER_TEXT;
```

src/components/Person/Person.jsx
```diff
@@ -1 +1,27 @@
...
+    <section className="Person">
+      <h2 className="Person__name">{`My name is ${name}`}</h2>
+
+      {age && <p className="Person__age">{`I am ${age}`}</p>}
+
+      {isMarried === true ? (
+        sexOfPartner
+      ) : (
+        <p className="Person__partner">I am not married</p>
+      )}
```
BudnikOleksii: - You don't need to compare with true as it's already boolean
- <p className="Person__partner"> is the same in all cases

BudnikOleksii (changes requested):
Good work 👍
Left a few suggestions

src/components/Person/Person.jsx
```diff
@@ -1 +1,20 @@
-// export const Person = ({ person }) => ();
+export const Person = ({ person }) => {
+  const { name, age, sex, isMarried, partnerName } = person;
+  const SEX = {
```
BudnikOleksii: You can move constant outside the component

BudnikOleksii (approved):
Good work 👍

PR #3080, opened 2025-05-13
task solution

RPrykhodko:
https://<RPrykhodko>.github.io/react_person/)


RPrykhodko:
https://.github.io/react_person/

PR #3097, opened 2025-05-14
Solution

meryam99:


meryam99:
@mateacademy-ai-mentor not all person have the age property according to the task

PR #3142, opened 2025-06-09
Solution

StronDator:
[DEMO LINK](https://StronDator.github.io/react_person/)


etojeDenys (changes requested):
The demo link doesn't work. Please fix it and then request a review again

PR #3456, opened 2025-11-07
task solution

iryna-knyzh:


iryna-knyzh:
@mateacademy-ai-mentor tag
 The task does not specifies using a gender property with values 'male' or 'female'

PR #3468, opened 2025-11-13
Solution

IrynaYurchak:
[DEMO LINK] https://IrynaYurchak/react_person


brespect (changes requested):
Sorry, I can`t reach your demo link, also try to pass all test before submit

PR #3587, opened 2026-02-07
Add solution

Kobeliukh:
[DEMO LINK](https://Kobeliukh.github.io/react_person/)


Kobeliukh:
Hello! I read your review. And I should pass an object as a single prop referring to this instruction:
 - Pass the entire person object as a prop using <Person person={misha} /> (do not use individual fields);
 
 Also, I can't use different formatting because of the linter. Thank you!
 
 @mateacademy-ai-mentor

PR #3622, opened 2026-03-09
solution

joselangoni:
[DEMO LINK](https://joselangoni.github.io/react_person/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import './App.scss';
 
+import { Person } from './components/Person/Person';
+
 export const misha = {
   name: 'Misha',
   age: 37,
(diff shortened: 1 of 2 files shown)


IanLuan (changes requested):
Great progress, you're almost there! Only two things are needed to follow the checklist:

1. **[CODE STYLE] - Use destructuring for getting access to values of props object**

Instead of using `person.name`, `person.age`, or any other attribute directly, you can destructure the `person` object into different constants, like this:

`const { name, age } = person`

Please do this for all attributes.

2. **[CODE STYLE] - Readabily is everything. Format ternary operator operands correctly - move each operand to the separate line:**

When using a ternary operator, instead of writing it on a single line, split it into multiple lines, like this:

```
{ person.sex === 'm'
    ? 'wife'
    : 'husband'
}
```

joselangoni:
@IanLuan I destructured the props object. But I can't move each operand of the ternary operator to a separate line, since it automatically returns to the same line.
I've already disabled 'format on save' and also 'Prettier', but after that it indicates a formatting error when separating ternary operands.
How should I proceed?

IanLuan (approved):
Great Job!

PR #3792, opened 2026-07-05
feat: updated person object to be passes as props

thorn086:
updated person object to be passed as props


thorn086:
updated comments

